PR #46, opened 2021-11-01
Incremental load fix

isc-svelury:
Git fetch and diff are now limited to the current branch in Pull(). 

IncrementalLoad now supports file deletions.


isc-svelury:
Will have to squash and merge

cls/SourceControl/Git/Utils.cls
```diff
@@ -342,15 +343,6 @@ ClassMethod Pull(remote As %String = "origin") As %Status
     set event = $classmethod(..PullEventClass(),"%New")
     set event.LocalRoot = ..TempFolder()
     merge event.ModifiedFiles = files
-    for i=1:1:$Get(files) {
```
isc-tleavitt, line 345: We still want to track modified items. Would be good to have data here if it's add/edit/delete.

isc-tleavitt: (This is a useful part of the pull event handler API.)

isc-tleavitt (changes requested):
One minor comment.

cls/SourceControl/Git/Utils.cls
isc-svelury: I think I am misunderstanding how this bit of code will work. 
After the merge, `event.ModifiedFiles` will have the same tree structure and nodes as `files`.
Then we iterate through the files, compute the internal name of the files and add a node for the iternal name to `event.ModifiedItems`. 

Once we call the IncrementalLoad event handler, it iterates through `event.ModifiedFiles` and again computes internal names for the external file names returned by `git diff` and merged into `event.ModifiedFiles`. It processes these files and then ignores the internal names computed in `Pull()` because we did not iterate over those indices. 

I think I am missing something here because from what I understand, this bit of code did not contribute to the end result. 

Are you saying we should add information on whether the commit adds, modifies or deletes the file so that we can have separate handling for each type of change? We name change the `--name-only` flag to `--name-status` for that.

isc-tleavitt: @isc-svelury I'd like the set of modified files and set of modified items (ideally with the operation - add/edit/delete) to be part of the API available to pull event handlers generally. Even though it's extraneous detail for the event handlers we have built in, there are use cases in applications of greater complexity that call for it.

isc-svelury: Okay, got it. 

I modified `event.ModifiedFiles` to store the corresponding internal name for each file. So the information is still available. 

Having it in `event.ModifiedItems` would make it clearer though. I will add that back in. I will also add the type of change to the `ModifiedItems`. I will construct `ModifiedItems` before the `git pull` because the merge might delete files and then their internal names would be an empty string. 

This information could be structured more clearly if we could use something like Python's dictionary or the JSON structure. Would be okay if I changed it to be:

```
set Modification("externalName") = externalName
set Modification("internalName") = internalName
set Modification("type") = type
set Modifications($I(Modifications)) = Modification
```

isc-tleavitt: @isc-svelury I think the cleanest approach would be either:
set modifications($i(modifications)) = $lb(externalName,internalName,action)
or using a registered object with ExternalName, InternalName, and Action properties as the value.

isc-svelury: Yup, that seems more like the ObjectScript way of doing things. Thank you!

isc-tleavitt: Technically this will be an API-breaking change but we're just going to call it a bugfix (since I'm sure nobody has written their own pull event handler in the past week...)

cls/SourceControl/Git/Modification.cls
```diff
@@ -0,0 +1,14 @@
...
+{
+
+/// path of the file
+Property externalName As %String;
+
+/// Name in IRIS SourceControl.Git.Modification
+Property internalName As %String;
+
+/// Type of change (A|C|D|M|R|T|U|X|B). See git diff documentation.
+Property changeType As %String;
```
isc-tleavitt, line 12: Longer-term, would be nice to have more of an API around this than using the single git character code. That'll be a follow-up.

isc-tleavitt (approved):

PR #60, opened 2021-11-05
Commit Explorer view now works for merge commits as well.

isc-svelury:
The merge-diff format is now separate instead of the default combined. This does not affect regular commits. Turned out to be a much simpler fix than I initially expected.


isc-tleavitt (approved):
Well, that was easy!

PR #61, opened 2021-11-05
Webui css to less

isc-svelury:
Ported over all the changes to git-webui.css to the .less file. After compilation there diff will show some differences but those are due to different grouping and/or positioning in the files.


isc-tleavitt (approved):
Great to have this done. Thank you!

PR #64, opened 2021-11-05
Standardize code style

isc-tleavitt:
Used AppS.CodeTidy with the new tab setting set to 4 spaces.

### DIFF
--- a/cls/SourceControl/Git/WebUIDriver.cls
+++ b/cls/SourceControl/Git/WebUIDriver.cls
@@ -35,38 +35,18 @@ ClassMethod HandleRequest(pagePath As %String, InternalName As %String = "", Out
             */
             if (pathStart = "git") {
                 if $piece(pagePath,"/",3) = "cat-file" {
-                    /*
-                    if self.path.startswith("/git/cat-file/"): {
-                    obj = self.path[14:]
-                    }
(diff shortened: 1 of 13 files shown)


isc-tleavitt:
Note, this doesn't work cleanly for the CSP files server-side. Not going to worry about those for now.

cls/SourceControl/Git/WebUIDriver.cls
```diff
@@ -5,144 +5,148 @@ ClassMethod HandleRequest(pagePath As %String, InternalName As %String = "", Out
...
         if (%request.Method = "GET") {
             set %data = ##class(%Stream.TmpCharacter).%New()
             // Things not handled from Python backend:
             /*
             def do_GET(self):
-                elif self.is_git_request():
-                    self.process_http_backend()
-                else:
-                    SimpleHTTPRequestHandler.do_GET(self)
+            elif self.is_git_request():
```
isc-tleavitt, line 31: Kind of annoying what happened to the commented-out python equivalent here (especially since spacing matters in that context)... Maybe it's just time to kill these comments.

isc-svelury: Do we still need this code? If so, we should go back and make an issue for it and embed the code in that. We can remove it from here then

isc-svelury (approved):

PR #65, opened 2021-11-05
Rotate refresh button

isc-svelury:
The refresh button now rotates while we wait for the git fetch --prune to run


isc-tleavitt:
Before merging - can you double check the behavior in Studio/IE? I feel like this won't work there, which is fine, but let's make sure it doesn't break things.

isc-tleavitt (approved):

isc-svelury:
It works properly in Studio and IE.

PR #66, opened 2021-11-05
Depends on git 2.31.0+

### LINKED ISSUE #62: Updated git version requirement for git diff --diff-merges
the --diff-merges option for git diff was added in git 2.31.0 (see https://git-scm.com/docs/git-diff/2.31.0 vs https://git-scm.com/docs/git-diff/2.30.0). I happened to be running an earlier version so the UI falls over now.

Two options:
* Detect version and avoid using the new flag on older versions
* Update documentation to say git 2.31.0+ is required

isc-tleavitt:
Fixes #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 ## Prerequisites/Dependencies
 
 * IRIS instance with license key
-* Git 2.x (tested as far back as 2.22)
+* Git 2.31.0+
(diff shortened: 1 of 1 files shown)


isc-svelury (approved):

isc-svelury:
Thank you for updating this!

PR #79, opened 2021-11-17
Git add packages

### LINKED ISSUE #71: Unable to add package path containing multiple objects (VSCode)
Attempting to add entire sub-package to git by right-clicking package folder on isfs object, selecting Source Control Commands and then Add. Get following error:

`fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths`

Right-clicking individual files (without opening) and doing the Add works fine.

isc-svelury:
Adding packages to source control now works as expected. Closes #71 

Known bug/potential improvement:

Currently, VS Code does not pass package names with the `.pkg` extension. So to check if a package name is passed, we look for classes in the package. If there are no classes in the package, then the package will not be recognized and this solution will not work in VS Code. It will still work on Studio though.


isc-tleavitt:
@isc-svelury this looks good. Can you file an issue report on vscode-objectscript re: consistency with studio in passing .pkg to extensions and mention this issue there?

PR #82, opened 2021-11-29
Queue filter

### LINKED ISSUE #28: Improved uncommitted queue + filter in web UI
A user should be able to commit:
* Files they have modified (e.g., uncommitted queue lists them as the user)
* Files they have deleted (need to track in uncommitted queue)
* Files modified without an owner listed (e.g., sc-list.txt)
* Files uncommitted with changes by another user AFTER confirming that they intend to

As part of the last point, the web UI should clearly display which files have uncommitted changes from other users. Files with uncommitted changes from the current user should be shown as they currently are.

It should also be possible to cancel another user's uncommitted change, but again with a prompt making sure the user really knows what they're doing.

isc-svelury:
1. Uncommitted Queue now tracks deleted files.
2. When returning uncommitted files,  the uncommitted API now also includes files that are tracked by git but not by the Uncommitted queue.
3. The WebUI  adds an icon to indicate that the file was edited by another user. 
4. For "available" files the process of staging, unstaging and cancelling changes remains unchanged for the user. 
4. When the user tries to act upon changes made by another user, a popup confirms that the user wants to take that action. 
5. Action of each "unavailable" file must be confirmed through its individual checkbox. 
6. If the user selects multiple files, the "available" files are acted upon in the background while the confirmation popup is active on the screen for the user to respond. Once the user confirms their choice (or cancels the action), processing resumes as normal.

This PR closes #28


git-webui/release/share/git-webui/webui/js/git-webui.js
```diff
@@ -1652,42 +1662,143 @@ webui.ChangedFilesView = function(workspaceView, type, label) {
         }
     };
 
-    self.getFileList = function(including, excluding) {
-        var files = "";
+    function confirmActionForUnavailalbleFile(files, action) {
```
isc-tleavitt, line 1665: Nit: typo (unavailable)

cls/SourceControl/Git/Extension.cls
```diff
@@ -248,6 +248,18 @@ Method IsInSourceControl(InternalName As %String) As %Boolean [ CodeMode = expre
 InternalName'="" && ##class(Utils).IsInSourceControl(##class(Utils).NormalizeInternalName(InternalName))
 }
 
+/// Called before an item is deleted.
+Method OnBeforeDelete(InternalName As %String) As %Status
+{
+    set context = ##class(SourceControl.Git.PackageManagerContext).ForInternalName(InternalName)
+    set InternalName = ##class(Utils).NormalizeInternalName(InternalName)
```
isc-tleavitt, line 255: Nit: full classnames are preferred (though other code in the same neighborhood might not be great about that)

cls/SourceControl/Git/WebUIDriver.cls
```diff
@@ -160,13 +160,17 @@ ClassMethod Uncommitted() As %SystemBase
         quit:key=""
         // Check that current user has files(key) uncommitted and only %Push if they do
         set filename = ##class(Utils).FullExternalName(key)
-        set sc=##class(SourceControl.Git.Change).GetUncommitted(filename,.tAction,.tInternalName,.UncommittedUser,.tSource,.UncommittedLastUpdated)
-        if $$$ISOK(sc) && ($data(tAction)&&(UncommittedUser=$username)) {
+        if (($ISVALIDNUM(key)) && (files(key) '= "")){
```
isc-tleavitt, line 163: Why do we need to check ISVALIDNUM?

isc-tleavitt (commented):
A few minor comments. Looks good code-wise; haven't actually tested the behavior but will do that before next release.

cls/SourceControl/Git/WebUIDriver.cls
isc-svelury: In `Utils.GitStatus()`, if we want to include all files, then for the files that don't have an internal name but are still tracked by git, I use `$increment` to generate keys. For files that have internal names, the internal names are the keys and the filepath is the value associated with the key. 

So in `WebUIDriver.Uncommitted()`, if the key is a number and the value is not empty, then it indicates that the file is tracked by Git but not by the Uncommitted queue. Therefore, any user should be able to take action on those changes.

cls/SourceControl/Git/Extension.cls
isc-svelury: This happens in multiple places all throughout the project. It might be better to create a separate PR for finding and refactoring all these instances.

git-webui/release/share/git-webui/webui/js/git-webui.js
isc-svelury: Fixed this.

isc-tleavitt (approved):

Issue #93, opened 2021-12-03
Extension on Other filetypes


CraigRegester:
See issue #92 for examples

HL7 schemas with `git-source-control` is outputting with a `.CLS` extension. `InterSystems ObjectScript extension` (see output above), outputs with `.hl7` like I would expect/prefer. 

Same scenario also happens with `LUT `files. I have both LUT and HL7 in my Mappings in Settings but it makes no difference - though does folder them like I want/specify in the mapping. Can we respect the original extensions and just use a .gitattributes file to instruct 'make them pretty' in GitHub (may not be necessary as they are XML)

```
*.cls linguist-language=ObjectScript
*.mac linguist-language=ObjectScript
*.int linguist-language=ObjectScript
*.inc linguist-language=ObjectScript
*.inc linguist-language=ObjectScript
*.csp linguist-language=Html

*.sh text eol=lf
*.cls text eol=lf
*.mac text eol=lf
*.int text eol=lf
Dockerfil* text eol=lf
```

isc-tleavitt:
@isc-svelury we should just always use the extension from the internal name.

isc-tleavitt:
Fixed in #100. Will need to get a new release out soon...

CraigRegester:
@isc-tleavitt echoing need for new release as last one was over 30 days now and I think you guys knocked out a ton of things (that would have been showstoppers for us) so excited to see if I can have others on our team start using this!

Appreciate you both, keep up the good work and thanks for all the efforts.

isc-tleavitt:
@CraigRegester done - published 1.2.0

PR #100, opened 2021-12-14
Preserve filenames for AbstractDocuments

### LINKED ISSUE #92: Dots (.) are being escaped automatically on Schema exports (and likely other files)
This is happening both with git-source-control and the InterSystems Objectscript Extension (not using ISFS).

If I have a schema named FAEPIC_2.3.1.HL7 and Add (ISFS) or Export (non-ISFS), I get this:

InterSystems ObjectScript Extension Export:
`export "FAEPIC_2.3.1.HL7" as "c:\Users\142990\MyDev\ISC-HealthShare-UCR\FALIB\src\FAEPIC_2\3\1.hl7" - Success 
`

```
InterSystems git-source-control Export:
new file:   INTGBUS/hl7/FAEPIC_2/3/1.cls
```

Can we not escape the dots on schema exports? Creates a nasty folder structure.

isc-svelury:
When nested folder structure is not enabled, files are now mapped into folders corresponding to their extension. No further folder nesting occurs.

To enable this behavior, turn off the folder switch in the settings. Note, this only works for a universal file type mapping. So the coverage must be *

This PR also fixes the bug where some file extensions were being changed to `.cls` incorrectly.

Closes #92 and #93

### DIFF
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -1239,24 +1239,32 @@ ClassMethod Name(InternalName As %String) As %String
             // Grab the actual name
             set actualName = $classmethod(docclass,"GetOther",InternalName)
             // The actualName is only valid if we get a single .cls as a result
-            if actualName'[",",$zconvert($piece(actualName,".",$length(actualName,".")),"U")="CLS" {
+            if (actualName'[",") && ($zconvert($piece(actualName,".",$length(actualName,".")),"U")="CLS") {
                 // We use what GetOther() gives us, thus Demo.Loan.FindRateProcess.bpl becomes Demo.Loan.FindRateProcess.cls
                 set InternalName = actualName
             }
         }
     }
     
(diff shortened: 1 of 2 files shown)


CraigRegester:
Not sure I quite follow the commit comment. Is this suggesting we can't folder the "Other" objects at all or that they need to be in one big folder? (E.g., .LUT files mixed with .HL7 files)

isc-svelury:
You will be able to folder "Other" objects. There is now a toggle added to each of the defined mappings in the settings UI to switch off the automatic nested folder creation that occurs when we split a filename by `.` . Once you switch off the toggle, the text box where you fill in the path should become a readonly field and have `NoFolders` filled in. 

This means that when you export those files out, the files will be stored in a folder corresponding to their extension (hl7/ for .hl7 files and so on) and no further nesting and/or directory creation takes place. So filenames for these files will be preserved.

CraigRegester:
Perfect, that makes sense now and what I hoped would be the case.

PR #101, opened 2021-12-16
Add Line break in Output window

CraigRegester:
Minor thing

In RemoveUncommitted, need to add line break after output as output is not cleared between subsequent actions - which is fine and likely preferred - but makes for a confusing run on sentence.

exporting new version of MyTable.lut to /hfs/sys/git/IRISHealth/INTGBUS/lut/MyTable.lut

/hfs/sys/git/IRISHealth/INTGBUS/lut/MyTable.lut removed from uncommitted list <Should be line break here> exporting new version of MySchema_2.5.1.hl7 to /hfs/sys/git/IRISHealth/INTGBUS/hl7/MySchema_2/5/1.hl7

/hfs/sys/git/IRISHealth/INTGBUS/hl7/MySchema_2/5/1.hl7 removed from uncommitted list

Note also the schema is still splitting into separate files on the dots. I thought this was fixed but may have misunderstood. Still looking at this part.

### DIFF
--- a/cls/SourceControl/Git/Change.cls
+++ b/cls/SourceControl/Git/Change.cls
@@ -34,7 +34,7 @@ ClassMethod RemoveUncommitted(FileList, Display = 1, Revert = 0, ActiveCommit = 
     set filename=FileList,SCs=$$$OK
     if $data(FileList)>1 set filename=$order(FileList(""))
     while (filename'="") {
-        if Display write !,filename_" removed from uncommitted list"
+        if Display write !,filename_" removed from uncommitted list",!
         if '..IsUncommitted(filename) set filename=$order(FileList(filename)) continue
         set obj=..OpenUncommitted(filename)
         if (obj.Source="trakcare")&&($data(^SYS("ChangeControlClass"))) {


isc-tleavitt:
@CraigRegester thanks!

isc-svelury:
Splitting by dots should be fixed. Did you toggle the NoFolders switch in the settings for that mapping?

CraigRegester:
> 

Yeah I was playing with that setting. If it is on (default) I get the old functionality (dot split, no good). So I toggle it off and now I get this weirdness:

```
cacheusr@oairsint1:/hfs/sys/git/IRISHealth$ ls
FITLIB                          INTGBUS                         NoFoldersMySchema_2.3.1.hl7
``` 

This is a bit of the convo we had yesterday maybe but what I was hoping we ended up at is I could still create top-level folders (I use it for namespace and hl7/ to keep everything tidy) but just not let the dots in the filename dictate additional folders.

I had this working in my "version" of SCM so I can find the relevant code snippet if interested and make a pull request to suggest the approach. Though the NoFolders string being prepended to the filename is definitely not desired.

image

isc-svelury:
You are right. I just sent out a fix for this. It did not show up when I was testing the solution initially and is a simple fix. 
After that gets pushed, it should work as expected. 

So, `MySchema_2.3.1.hl7` should be exported to `hl7/MySchema_2.3.1.hl7`. 

Apologies for the back and forth on this.

CraigRegester:
No worries or need to apologize to me, much appreciated!

PR #108, opened 2022-01-11
Uncommitted check on checkout

### LINKED ISSUE #107: Uncommitted changes owner check for switching branches
Before switching branches, we should check if all the uncommitted changes were made the current user. If not, we should inform the user and confirm the action like we do for staging, stashing, unstaging and discarding changes. We should direct the user to the workspace so that they can either stash, stage or cancel other users' changes.

isc-svelury:
Closes #107 . The UI calls the `uncommitted` API and checks if any of the modified files were changed by a different user than the one signed in. 
- If yes, then a dialog pops us asking the user to warn the user that they will be carrying over those changes as well and confirms that the user wants to go ahead with the branch switch. 
- If no, then checkout proceeds as before.

Works for local and remote branch checkout, and branch creation. 

Note: There is a lot of repeated code/code structure in this PR. I first tried to wrap the popup creation in a different function but the if-else blocks that execute the appropriate action have to be different for different use cases. So then I just put in the code to determine the flag variable into a different function but because JS is asynchronous, the function is called and the if-else block get executed and then the flag variable is set. I tried a few things after that to make the execution more "synchronous" but each method seemed more janky and sub-optimal than just repeating the code everywhere it's needed. I am interested to see if you know of a way to get around this or if you have a different way to reduce the amount of repeated code.

### DIFF
--- a/git-webui/release/share/git-webui/webui/css/git-webui.css
+++ b/git-webui/release/share/git-webui/webui/css/git-webui.css
@@ -47,14 +47,17 @@ body {
   font: 10pt sans-serif;
 }
 #confirm-branch-delete,
-#confirm-staging p {
+#confirm-staging,
+#confirm-branch-checkout p {
   font-size: 105%;
 }
 #confirm-delete,
 #cancel-delete,
 #confirm-force-delete,
 #confirm-staging,
-#cancel-staging {
+#cancel-staging,
+#confirm-checkout,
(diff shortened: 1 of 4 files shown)


isc-tleavitt (approved):

isc-svelury:
Merged files locally. Merging PR now.

PR #125, opened 2022-02-16
No more sclist

### LINKED ISSUE #120: Track resources in the repo via global rather than sc-list.txt
As part of this:
Import should be made package manager-aware.
Import in a non-package manager setting should just do a $System.OBJ.ImportAll() on all folders listed in the mappings global for various types.

@isc-svelury you've been in this code more than I have - do you see use cases for sc-list.txt that I'm missing (where we couldn't just track the list of files in a global)?

isc-svelury:
1. Remove all references to sc-list from codebase. 
2. Remove workaround for `.pkg` not being added in VS Code. This was addressed here: https://github.com/intersystems-community/vscode-objectscript/pull/776
3. Add and Remove are now both displayed when the user selects the Server Source Control Menu for a package. 
4. We are no longer concerned with checking if a package should be added to source control (if all its files are in source control) and removing a packaging from source control (if even one of its files is no longer in source control). Packages are never directly added to source control, only their files are. 
5. Created 2 new methods to add and remove files from server side source control only. These are used after git pull, before calling the pull event handler. 
6. Removed special handling for deleted files from the `IncrementalLoad` event handler. 

Closes #120


isc-tleavitt:
This looks good and relatively straightforward to me from a code standpoint. I'm going to squash and merge, but will want to do some testing around this specifically before our next release. (But I'll be testing everything, not just this, so makes sense to merge in the meanwhile.)

Issue #146, opened 2022-03-14
GitStatus() does not filter out empty strings


isc-svelury:
The method stores empty strings in the `files` variable that is passed by reference. 

Example:

```ObjectScript
USER>do ##class(SourceControl.Git.Utils).GitStatus(.files, 1)
 
USER>zw files
files=4
files(1)=$lb("??","notes.md")
files(2)=$lb("??","nspace/hl7/2.1.1.1.1.hl7")
files(3)=$lb("??","nspace/hl7/TestSchema.hl7")
files(4)=$lb("","")
files("UnitTest.SourceControl.Git.IncrementalLoadTest.CLS")=$lb("??","test/UnitTest/SourceControl/Git/IncrementalLoadTest.cls")
```

PR #148, opened 2022-03-17
Strip leading whitespace to type of change, filter out empty rows

isc-svelury:
The `GitStatus()` method does not filter out cases where the external name is an empty string, and leaves in some leading whitespace in the type of change. I fixed these issues.


isc-svelury:
Closes #146

PR #157, opened 2022-03-21
Graph fill fix

### LINKED ISSUE #141: Graph flashes and fills the whole page
The graph flashes and fills the whole commit view sometimes. This can be seen on VS Code if the number of local branches in more than 5. It only seems to appear on page load and disappears if I switch to a different tab and back to the commit history.

isc-svelury:
I think the issue was happening because, on load, we were trying to display the log of ref before all the async git requests were finished but I am not a 100% sure. 

Moving the `selectRef` to after we finish building the sidebar seems to do the trick though. 

Closes #141

### DIFF
--- a/git-webui/release/share/git-webui/webui/css/git-webui.css
+++ b/git-webui/release/share/git-webui/webui/css/git-webui.css
@@ -26,7 +26,7 @@
 }
 .refresh-start {
   animation-name: rotate;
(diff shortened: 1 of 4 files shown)


isc-tleavitt (approved):
This makes sense to me. Seems like a tricky one!

PR #158, opened 2022-03-21
No more bifrost

### LINKED ISSUE #140: Graph creates many unnecessary lines after the first commit.
Graph glitch below the first commit:
image

isc-svelury:
An invalid entry was being added to the log with the whole git log. I rewrote the conditions to catch that and make sure the invalid entry isn't appended. 

Closes #140


isc-tleavitt (approved):
Best PR title ever.

PR #163, opened 2022-03-22
Add input validation to settings

### LINKED ISSUE #150: Settings' form fields have no input validation
Mappings should all be required fields. 

We already verify that the git bin exists at the path provided but we can display the confirmation to the users better. 

We should also verify that the SSH key exists.

isc-svelury:
1. Git bin path, path of the SSH key and mappings.
2. I also took the opportunity to change the text on the tooltips of the `NoFolders` toggle depending on its state. \

Closes #150


isc-tleavitt (commented):
One question here

Issue #130, opened 2022-02-17
Output of Schema with periods (dots) in name breaks SC Ad


CraigRegester:
(latest commit testing)

When adding a Schema to source control (SC Menu->Add), receive error in browser:

![image](https://user-images.githubusercontent.com/65192819/154551136-dce93298-6965-4fb2-9e77-ec992bdd5fb3.png)

Looking in WebUI afterwards, I can see it partially worked but clearly does not like the dots - note if I click either of the items in this screenshot, I see nothing.

![image](https://user-images.githubusercontent.com/65192819/154551282-c8e639cf-00df-4be5-9b68-488895274f83.png)

Maybe has something to do with the output pathing I desire to use (which you can see on 2nd screenshot.)

CraigRegester:
So that's the 2nd part of the issue then but seems like the first part (where it's referencing the new global for tracking) is different and also out it outputs on the filesystem... it's creating folders w/ the dots still.

Output in the filesystem:

```shell
In_Cdi_Rslt_2/3:
total 8
drwxrwsr-x    2 cacheusr gitusers        256 Feb 17 13:50 .
drwxrwsr-x    3 cacheusr gitusers        256 Feb 17 13:50 ..
-rwxrw-r--    1 cacheusr gitusers        339 Feb 17 13:50 1.hl7
```

isc-svelury:
From my understanding it looks like the file isn't being exported as expected and so the` InternalName` is coming up as an empty string. I think that's where we get the subscript error is from. 

How are your mappings configured? I am not able replicate the issue. I switched the mapping to `.inc` files to NoFolders and adding and removing from source control seems to work as expected for me. The flat folder structure is maintained even when I add multiple `.inc` files from different packages.

CraigRegester:
Yea so if I turn on 'NoFolders', it "works" to a degree (though I still got the subscript error):

![image](https://user-images.githubusercontent.com/65192819/154566816-1b77c580-1f3e-46c0-b1e4-b38f8095f0c3.png)
![image](https://user-images.githubusercontent.com/65192819/154566878-6ec86819-3909-4220-a0e0-6882fe0bd7f3.png)

Wonder if I need to flush out git-source-control and reload (as I tried to just load the latest repo code using zpm without purging out the previous stuff... so did it initialize anything you need to track the globals properly after switching from sclist?

It does output "right" filename wise with NoFolders

```shell
[...]

PR #167, opened 2022-03-22
NoFolders check for mappings in ListItemsInFiles()

### LINKED ISSUE #162: ListItemsInFiles() does not work with NoFolders
When a mapping has `NoFolders` enables, the file path is stored under the `NoFolders` subscript rather than the mapping coverage subscript. `ListItemsInFiles()` does not take this into account and throws an error.

isc-svelury:
Closes #162

### DIFF
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -1055,8 +1055,11 @@ ClassMethod ListItemsInFiles(ByRef itemList, ByRef err) As %Status
         set mappingCoverage = $order($$$SourceMapping(mappingFileType, ""))
 
         while (mappingCoverage '= ""){
-
-            set mappedRelativePath = $$$SourceMapping(mappingFileType, mappingCoverage)
+            if ($data($$$SourceMapping(mappingFileType, mappingCoverage)) = 1){
+                set mappedRelativePath = $$$SourceMapping(mappingFileType, mappingCoverage)
+            } else {
+                set mappedRelativePath = $$$SourceMapping(mappingFileType, mappingCoverage, "NoFolders")
+            } 
             set mappedFilePath = ##class(%File).NormalizeFilename(mappedRelativePath, ..TempFolder())
             
             if (##class(%File).DirectoryExists(mappedFilePath)){
@@ -1603,4 +1606,3 @@ ClassMethod GetSourceControlInclude() As %String
 }
 
 }
-


cls/SourceControl/Git/Utils.cls
```diff
@@ -1055,8 +1055,11 @@ ClassMethod ListItemsInFiles(ByRef itemList, ByRef err) As %Status
         set mappingCoverage = $order($$$SourceMapping(mappingFileType, ""))
 
         while (mappingCoverage '= ""){
-
-            set mappedRelativePath = $$$SourceMapping(mappingFileType, mappingCoverage)
+            if ($data($$$SourceMapping(mappingFileType, mappingCoverage)) = 1){
+                set mappedRelativePath = $$$SourceMapping(mappingFileType, mappingCoverage)
+            } else {
+                set mappedRelativePath = $$$SourceMapping(mappingFileType, mappingCoverage, "NoFolders")
```
isc-tleavitt, line 1061: Has the value at the "NoFolders" subscript changed from a flag to a directory? (This would actually explain something I was confused about before.)

isc-tleavitt (commented):
Looking for clarification here - I may have misunderstood previous changes. "NoFolders" used to just be a boolean flag and it seems like it changed to a directory?

cls/SourceControl/Git/Utils.cls
isc-svelury: It has. So that we can use a path with `NoFolders` mappings as well and discussed here: https://github.com/intersystems/git-source-control/issues/130#issuecomment-1043432587

Instead of having 2 entries, I just put both under one.

isc-tleavitt: Got it. My main concern is that this is backwards-incompatible; old configuration won't work properly. Why can't we have the flag and folder separate?

isc-tleavitt: To be clear: I think it's cleaner to keep the flag and folder separate.

isc-tleavitt: Actually thinking about this more I'm not sure what I think. Regardless I think we should automate the update from the old settings version to the new settings format on package update. (This will be done from an <Invoke> in module.xml.)

isc-svelury: At the time I thought there would be less confusion if they were in the same entry. I don't think that will be the case though. Should I change it to be in two separate entries? I can do that in a different PR and we can just close this one because we need to reevaluate if this issue will still be relevant at that point.

isc-svelury: To be backwards-compatible, the only thing we have to change in the settings is to add the extension as the path.

isc-tleavitt: Let's change it back to be two separate entries. Let's cancel this PR and reevaluate after splitting back out.

PR #168, opened 2022-03-23
Separate NoFolders mappings

isc-svelury:
As discussed in #167, I separated the `NoFolders` mappings into two entries in the global.


isc-tleavitt (approved):

isc-tleavitt:
Thanks, merged; is #162 still relevant?

PR #171, opened 2022-03-23
In RefreshUncommitted(), remove item from uncommitted queue if there is no change according to git

### LINKED ISSUE #155: RefreshUncommitted() should also use git status
`RefreshUncommitted()` does not remove files that are still in the Uncommitted queue but are not in the Git status.

isc-svelury:
If you do commits from the WebUI, the Uncommitted queue keeps accumulating records of changed files which don't get cleared out. This way, there is a way to clear out old changes that have been committed but their records haven't been removed. 

Ideally we want to remove the record when the change is committed though. 

Closes #155


isc-tleavitt:
looks good.

PR #174, opened 2022-03-24
Unit tests for Name() and Add granularity for NoFolders mappings

### LINKED ISSUE #154: Unit test for Name()
Given the `InternalName` of a file, `Name()` returns the path of the file relative to the repository root. We need to have unit tests for this.

### LINKED ISSUE #173: NoFolders for specific packages
We can provide more granularity with the way the `NoFolders` setting works. We should support:

1. Keeping all packages for a file-type foldered except some. 
2. Have no folders for all files of a filetype except certain packages which are foldered. 

For a given file, if no mapping match can be found, the default behavior is to create a folder "<file extension>/" and store the file there without splitting by ".". This will not be changed.

isc-svelury:
1. Added test cases for the `Name()` method
2. Added granularity in the way mappings work as described in #173 

Closes #154 and closes #173


isc-tleavitt:
Looks good!

PR #212, opened 2022-12-02
Support pre-IRIS installation

### LINKED ISSUE #31: Support installation pre-IRIS w/o package manager
We want to support 2016.1+ (maybe 2016.2+ for simplicity if JSON ends up mattering)
Should also cover generation of some artifact that can be installed on IRIS instances without using the package manager (maybe w/ Studio Project packaging? though that'll be challenging for the web resources)

isc-tleavitt:
Fixes #31 

This is still a WIP, but putting it up for comment in advance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.6] - 2022-12-06
+
+### Added
+- Support for Caché/Ensemble 2016.2.3, 2017.1.2, 2017.2.1, 2018.1.0 and later. (Treated as patch version bump because no compatibility impact for existing users.)
+
+### Changed
+- Various minor things under the hood to support use without the package manager and/or on older platforms
+
 ## [2.0.5] - 2022-12-02
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


isc-tleavitt:
Current blockers:
* I'm having trouble getting a functional Ensemble instance installed to test on - having weird private webserver issues. Going to try again with a different kit Monday.
* SourceControl.Git.Change is a subextent of %Studio.SourceControl.Change which isn't allowed pre-IRIS because %Studio.SourceControl.Change is (or at least SHOULD be) in a read-only database. Rules on subextents changed in IRIS.

Draft artifact in case somebody sees this and wants to give it a try - just extract the .xml and `do $system.OBJ.Load(that file,"ck")`:
[GitSourceControl_CE.zip](https://github.com/intersystems/git-source-control/files/10144166/GitSourceControl_CE.zip)

isc-tleavitt:
Another issue, namespace change on gitprojectsettings.csp doesn't work automatically and will need to change a bit to work pre-IRIS. (Probably a difference in the default template for /isc/studio/usertemplates.)

isc-tleavitt:
[GitSourceControl_CE.zip](https://github.com/intersystems/git-source-control/files/10169951/GitSourceControl_CE.zip)

Updated artifact attached.

isc-tleavitt:
@isc-pbarton I'll want to add docs on producing + loading the artifact as well before publishing... in short, it's just:
do ##class(SourceControl.Git.Utils).BuildCEInstallationPackage("GitSourceControl_CE.xml")

Then:
do $System.OBJ.Load("GitSourceControl_CE.xml","ck")

isc-pbarton (approved):
Tested this out on my Cache 2018.1.7 instance, looks to be working.

isc-tleavitt:
Committing this as-is, will doc and automate later.

Issue #217, opened 2022-12-29
"Import All" doesn't pick up newly-added files


isc-tleavitt:
It checks to see if they're in source control, which of course they aren't. (It just shouldn't check this.)

While investigating, also noted some issues operating in a package manager context. "Import All" and "Export All" should apply only to files in the current package context (or possibly the current lack thereof).

a-schuetz:
Maybe instead of skipping the check you could add the new files to source control like this:

```Objectscript
#dim sc as %Status
if '..IsInSourceControl(internalName) {
    set sc = ..AddToServerSideSourceControl(internalName)
} else {
    set sc = ..ImportItem(internalName, force)
} 
```

Im not sure if this causes any unwanted behaviour. But for example if a checkout a remote branch wich contains new files and import it into IRIS, those files are already tracked by the Git repository, so it would make sense to also add them to the items in the global.

isc-tleavitt:
@a-schuetz thanks for the comment. There are a few different paths where a new file could show up (including import) and really the global should be updated by any of them. I'm also not sure about possible unwanted behavior but will give that a try. Also, if you feel like digging around the code and submitting a PR you're always welcome - though this probably isn't a great introductory issue.

a-schuetz:
I created a PR #218. I did test use cases that I had in mind and it works as I would expect it to. But maybe I didn't cover all use cases.

PR #218, opened 2023-01-03
Fix import ignores newly-added files

a-schuetz:
Newly added files are imported to server side source control instead of ignored.

### DIFF
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -1138,10 +1138,12 @@ ClassMethod ImportRoutines(force As %Boolean = 0) As %Status
     for  {
         set internalName = $order(itemList(internalName))
         quit:internalName=""
-        if '..IsInSourceControl(internalName) {
-            continue
+        #dim sc as %Status
+        if ..IsInSourceControl(internalName) {
+            set sc = ..ImportItem(internalName, force)            
+        } else {
+            set sc = ..AddToServerSideSourceControl(internalName)
         }
-        #dim sc as %Status = ..ImportItem(internalName, force)
         if $$$ISERR(sc) {
             set ec = $$$ADDSC(ec, sc)
         }


cls/SourceControl/Git/Utils.cls
```diff
@@ -1138,10 +1138,12 @@ ClassMethod ImportRoutines(force As %Boolean = 0) As %Status
     for  {
         set internalName = $order(itemList(internalName))
         quit:internalName=""
-        if '..IsInSourceControl(internalName) {
-            continue
+        #dim sc as %Status
+        if ..IsInSourceControl(internalName) {
+            set sc = ..ImportItem(internalName, force)            
+        } else {
+            set sc = ..AddToServerSideSourceControl(internalName)
```
isc-tleavitt, line 1145: Are such newly added files actually imported? Looking at the diff I'd think not, but maybe this happens deeper down - haven't actually tested. Presumably we'd always want to call ImportItem?

isc-tleavitt (commented):
One note here re: testing.

Note, I have a fix in progress that covers some semi-related issues too. (Should have explicitly stated that.) Feel free to take a look at the import-export-fixes branch.

a-schuetz:
fixes #217

cls/SourceControl/Git/Utils.cls
a-schuetz: Yes, `AddToServerSideSourceControl` first updates the global and than calls `ImportItem`.

```Objectscript
ClassMethod AddToServerSideSourceControl(InternalName As %String) As %Status
{
     #dim i as %Integer
    #dim ec as %Status = $$$OK
    for i = 1:1:$length(InternalName, ",") {
        #dim item as %String = ..NormalizeExtension($piece(InternalName, ",", i))
        set @..#Storage@("items", item) = ""
        #dim sc as %Status =  ..ImportItem(item, 1)
        if 'sc {
            set ec = $$$ADDSC(ec, sc)
        }
    }
    quit ec
}
```

isc-tleavitt:
@a-schuetz I've folded your change into a larger overhaul of import/export that fixes a bunch of other issues too. See #220. As such, I'm closing this PR, but thank you for your contribution and for reporting the issue in the first place!

PR #221, opened 2023-01-13
Add build/test/pre-IRIS-package script

### LINKED ISSUE #52: Set up CI and expand unit test library
Would be good to have tests run on both Windows and Linux if possible. (If not, leaving to devs to run on Windows and running automatically on Linux would be next best.)

isc-tleavitt:
Fixes #52 (could always do more for unit tests, but this is the first part of the battle).

Also turns out a bit of work is needed to get unit tests running on a container when git-source-control isn't loaded in development mode (vs. on an instance on a Windows machine where it is).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,7 +88,7 @@ jobs:
         if: always()
         with:
           name: "PreIRISInstallationPackage"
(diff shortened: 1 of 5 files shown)


isc-tleavitt:
@isc-pbarton, I would appreciate your review on this - pay no attention to the ridiculous number of commits trying to get the workflow working. (That's how it goes!)

isc-pbarton (approved):

PR #233, opened 2023-01-24
Improve non-https behavior

### LINKED ISSUE #227: Improve behavior over non-https connection
Without this, VSCode webviews just won't work. (And they fail badly.)

Could also perhaps make this clearer in the settings / git webui pages themselves. Specifically: if you attempt to access one of these pages on non-localhost over an http connection, offer a link that opens in a new window and yell at the user for not using https.

Better yet: detect menu listing over Atelier APIs with insecure connection, yell at the user right in VSCode, and automatically open in a browser window (Action=3) rather than an iframe if it's not going to work.

isc-tleavitt:
Fixes #227

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Link from WebUI to Settings page works properly (#230)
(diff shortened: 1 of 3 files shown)


isc-pbarton (approved):
Looking for documentation on the UserAction values, the first result is a Developer Community answer by (of course) tleavitt: https://community.intersystems.com/post/server-side-source-control-hooks-and-atelier-useraction3-not-working

PR #242, opened 2023-02-08
Fix behavior with BPL/DTL editing via Studio

### LINKED ISSUE #241: Changes in BP Designer cannot be commited when using IRIS Studio
When changes to a BPL are made via the Business Process Designer in IRIS Studio and saved, those changes are not replicated to the Git folder in the filesystem. Therefore the changes can't be commited. When the BP Designer is used via Browser in the Management Portal it works fine.

A workaround in IRIS Studio is to remove the file from SC and add it again. Then changes will be synced to the folder once.

Steps to reproduce:
1. Open a BPL in IRIS Studio (Workspace -> Business Processes)
2. Make any change to a BPL that is already added to SC (e.g. add an annotation) and click the save button.
3. Try to commit the change. You will also notice that the change has not beend synched to the .cls file in the Git folder.

isc-tleavitt:
Allow various places to change InternalName to the actual InternalName we're working with

Fixes #241

### DIFF
--- a/cls/SourceControl/Git/PackageManagerContext.cls
+++ b/cls/SourceControl/Git/PackageManagerContext.cls
@@ -34,6 +34,7 @@ ClassMethod ForInternalName(InternalName As %String = "") As SourceControl.Git.P
 {
     set instance = ..%Get()
     set instance.InternalName = InternalName
+    set InternalName = instance.InternalName
     quit instance
 }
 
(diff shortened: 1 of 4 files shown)


isc-tleavitt:
Previously, editing DTL through Studio would result in the URL to the editor being exported into the file.

cls/SourceControl/Git/PackageManagerContext.cls
```diff
@@ -34,6 +34,7 @@ ClassMethod ForInternalName(InternalName As %String = "") As SourceControl.Git.P
 {
     set instance = ..%Get()
     set instance.InternalName = InternalName
+    set InternalName = instance.InternalName
```
isc-pbarton, line 37: Nit: InternalName parameter should be marked as ByRef

isc-pbarton (approved):

PR #247, opened 2023-02-15
Fix unexpected ZPM impact

isc-tleavitt:
ZPM reports the installer class via GetOther which was messing up package manager awareness for the git repo root. (Really convoluted how this ended up happening.)

This code is intended for BPL/DTL only so it makes sense to enforce that.

This is a fix to unreleased changes, so not making a note in CHANGELOG.md.


isc-tleavitt:
Planning to add a unit test for this too.

isc-tleavitt:
@isc-pbarton I don't think I'll have time to add a test for this before going on leave. Would appreciate your review as-is so I can get a new release out before then.

PR #281, opened 2023-09-05
Added configurable flag to make settings UI read-only

### LINKED ISSUE #258: Hide settings page (or make it RO)
Add ability to either hide settings page OR make it RO.

isc-pbarton:
resolves #258


isc-tleavitt (commented):
I'd argue that we should only make top-level settings read-only, and permit editing of user-level settings (git username/email).

@eduard93 would you agree?

isc-tleavitt:
@isc-pbarton see my earlier comments here (at your convenience...)

isc-tleavitt (approved):

PR #283, opened 2023-09-13
ZPM load in dev mode will build the javascript application

### LINKED ISSUE #37: Automate npm install, grunt release for load -dev
To support our minimal UI build process. Need to deal with #36 first to avoid causing issues.

isc-pbarton:
Resolves #37

### DIFF
--- a/cls/SourceControl/Git/Build.cls
+++ b/cls/SourceControl/Git/Build.cls
@@ -8,10 +8,10 @@ ClassMethod BuildUIForDevMode(devMode As %Boolean, rootDirectory As %String)
     }
     write !, "In developer mode, building web UI:"
     set webUIDirectory = ##class(%File).SubDirectoryName(rootDirectory, "git-webui")
-    write !, "npm install"
-    write !, $zf(-100, "/SHELL", "cd", webUIDirectory, "&&", "npm", "install")
+    write !, "npm ci"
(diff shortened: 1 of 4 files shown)


cls/SourceControl/Git/Build.cls
```diff
@@ -0,0 +1,18 @@
...
+
+ClassMethod BuildUIForDevMode(devMode As %Boolean, rootDirectory As %String)
+{
+    if 'devMode {
+        return
+    }
+    write !, "In developer mode, building web UI:"
+    set webUIDirectory = ##class(%File).SubDirectoryName(rootDirectory, "git-webui")
+    write !, "npm install"
+    write !, $zf(-100, "/SHELL", "cd", webUIDirectory, "&&", "npm", "install")
```
isc-tleavitt, line 12: This should be npm install --prefix webUIDirectory rather than using cd. (Same below, npm run --prefix)

isc-tleavitt (changes requested):

cls/SourceControl/Git/Build.cls
isc-pbarton: At least for me, `npm install` does not respect `--prefix`. But `npm ci --prefix` does work.

isc-tleavitt (approved):

isc-tleavitt:
As this is internal I don't think we need anything in changelog.md but feel free to add if you disagree.

PR #292, opened 2023-10-11
Private memory store

isc-rsaptars:
Added a private memory store utility


isc-etamarch:
Want some initial review for the work. 

My biggest concern is that we are setting the remote to be https://oauth2:<token>@<remote-url>, and thus having the token out in the open counteracts all of the effort to store it safely in memory. 
-> My first thought here is maybe to set / unset the remote for every git command, though this would probably drastically increase runtime (tripling amount of git commands for each command)

isc-tleavitt:
One more thing - add opt-in to work around case of hardcoded access token in URL, then mark "Ready for review"

PR #306, opened 2023-12-14
Fix for git submodules support

### LINKED ISSUE #305: Add support for git submodules
I use modules for different namespaces in my project setup as git submodules
Consider the following directory tree:

```.
├── .git/ //main git directory
├── .gitmodules
└── modules/
    ├── ModuleA/
    │   ├── .git //git submodule A
    │   └── module.xml
    └── ModuleB/
        ├── .git //git submodule B
        └── module.xml
```

The problem I suppose is that sources are not correctly mapped because the package manager is looking for a **directory** named ".git" at the root of the module folder, however submodules are just simple files linking to the parent git directory

according to this line:
https://github.com/intersystems/git-source-control/blob/9aa8009111a14328118e0dcfef02cd7b3101a1e6/cls/SourceControl/Git/PackageManagerContext.cls#L27C1-L27C166

exemple .git file in a submodule:

```
$ cat modules/ModuleA/.git 
gitdir: ../../.git/modules/modules/ModuleA
```

Antoine-dh:
resolves #305

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.1] - Unreleased
+
+### Fixed
+- Support for git submodules in package manager-aware setting (#305)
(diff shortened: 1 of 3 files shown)


isc-tleavitt (approved):

isc-tleavitt:
@Antoine-dh thank you!

Issue #315, opened 2024-01-09
Studio (and vs code) performace


EnricoParisi:
Since I started testing git-source-control I noticed a severe performance impact in Studio.
Before reporting I wanted to make some analysis to provide some more detail other then "it slows Studio down".

Opening an item like a class or csp page is much slower when git-source-control is used, for example, opening studio with 10 classes take around 10 seconds.
Even closing Studio is slow when you have some class opened.

In my test case I have a total of 2285 items (classes and csp pages) defined in git-source-control and it is clear that more items you have defined, the slower Studio gets.

I've been trying to investigate the issue and I realized that the implementation is very, very, very....inefficient.

Just as en example here are a few measurement/call count that the implementation perform  opening Studio with 10 classes:

32 calls to the ##class(SourceControl.Git.Utils).GitStatus() method. Note that this method call git.exe using $zf(-100), therefore is not "lighting fast".

40000+ calls to the ##class(SourceControl.Git.Utils).Type() method. I understand that this method "only" perform 78 global references and 1250 commands, bu calling it 40k+ times for opening 10 classes it seems a bit too much to me.

The worst problem is that after Studio has opened, it becomes very tedious to work with it, every action like moving the cursor, searching, typing is lagging, virtually impossible to comfortably work with it.

I think the issue is that Studio calls every few (5?) seconds the method ##class(SourceControl.Git.Extension).OnBeforeTimestamp(), this methods takes around 3.5 seconds to execute and perform 9'890'575 commands and 416'347 global references....every few seconds!

I haven't fully analyzed the vscode implementation, however from a preliminary test it seems a similar situation is present, apart from the lagging when editing that does not happen in vs code. In vscode it seems that the lagging happen when opening a class, changing active tab/class and performing server side operations like save/compile.

Bottom line, the current implementation is not suitable for any complex (I would say more that 500 items) environments.

Has anybody ever tested git-source-control with a significant (i.e. real world) number of items using Studio?

isc-tleavitt:
@EnricoParisi I appreciate the analysis on this - see #269 which has some of the same. This is very much on our radar for the near term (this month).

isc-tleavitt:
The fix boils down to a few different things, I think:
* Caching results in SourceControl.Git.Change:RefreshUncommitted and throttling calls to it (meaning, we call GitStatus less frequently)
[...]

Issue #269, opened 2023-07-28
Studio: hanging on Git operations, Git menu very slow to load


zdateh:
I am unable to practically use the Git menu in studio - doing operations like 'Fetch from Remote' just hangs Studio, with no error or anything in the console. Are there any ways I can diagnose what is causing this?

The Git menu is also slow to load - it takes around 5 seconds to display, which makes it pretty hard/annoying to use whilst working. I also fancy this extension has slowed down Studio in general.

By using a combination of what works and Git bash, I can do change control, but not really ideal!

isc-tleavitt:
@zdateh are you set up with an https remote? This sounds like https://github.com/intersystems/git-source-control/issues/235.

zdateh:
@isc-tleavitt Hi, yes we're using an https remote (Azure dev ops). Would ssh remotes potentially help (that's the other option provided in Azure)?

re: #235 I haven't seen any credential checks coming up

isc-tleavitt:
@zdateh yes, switching to ssh would fix the issue.

The problem is that the credential check is coming up in the background and hanging the process on the server. :(

isc-tleavitt:
Closing this as a duplicate of #235

zdateh:
Switching to SSH has indeed fixed my remotes issue - all git operations appear to succeed so far, thanks!

However it's still as slow in studio as before :( I am unsure what it could be doing in the background as git operations are near-instant from the command line, but as soon as I click on that Git menu, it's doing nothing for 5 seconds. Any ideas?

zdateh:
Hi, could we please re-open this? We have been using SSH as requested but the menus and studio operation is still noticeably slower with Git extension active. e.g.:

- re-opening/closing studio with many routines/classes being edited takes over a minute.
- loading a routine/class takes 5-10 seconds

None of them dealbreakers but can be annoying whilst developing :)

Cheers

isc-tleavitt:
@zdateh sure, happy to reopen given your observations.

Technically speaking (for whoever picks this up):
We run "git status" to see what's currently pending. There's some caching of the results built-in for rendering the menus (so that we don't re-run for each menu item), but not across operations (e.g., loading multiple tabs with different files in Studio). Load of a single class/routine taking 5-10 seconds is particularly unexpected.

@zdateh - if you could capture %SYS.MONLBL (RtnLine/Time/TotalTime) on SourceControl.Git.* and attach as CSV that would be super helpful.

zdateh:
[...]

PR #255, opened 2023-05-11
Simplify pull and configuration

isc-tleavitt:
(Details in changelog.)

Fixes #234
Fixes #237
Fixes #33
Fixes #254
Fixes #229

PR #318, opened 2024-01-12
Various performance fixes

isc-tleavitt:
Throwing this out there as a draft - more testing is needed (both cross-OS on Windows/Linux AND cross-version with/without /ENV on $zf(-100)) before saying it's good to go.

This address both #315 and #269, as well as some issues I've seen with lag in the WebUI, by the following measures:
* Caching the InternalName/ExternalName (that is, in the DB + on the filesystem) mapping
* Caching results of %Library.RoutineMgr:UserType
* Removing needless calls to %File:NormalizeFilename (which is surprisingly slow)
* Removing /SHELL from $zf(-100) calls as much as possible
* Reducing WebUI calls to RefreshUncommitted and in general putting a 10-second throttle on refreshes of the uncommitted queue (which may be bypassed for things like e.g. commit)

Outcomes for a larger project are:
* Opening a Studio tab (or 10) is much faster
* The WebUI goes from laggy to virtually instantaneous, most notably for the initial load but also with a significant improvement when actually viewing uncommitted changes

Testing should cover:
* Ensuring no regressions with the old issue trying to use root's global attributes on Linux (without /SHELL, covered using $XDG_CONFIG_HOME with /ENV as described at https://git-scm.com/docs/git-config#FILES) - need to test on versions without /ENV support in $zf(-100)
* General git operations on Windows/Linux (unit tests are actually pretty representative here for the Linux side, and I've run locally on Windows)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Deletion of files in locked environment is now suppressed (#302)
 - Failed to import file VS Code popup no longer shows up after overwriting file on server once (#264)
 - Don't automatically stage files added to source control (#303)
+- Performance improvements (#269, #315)
 
 ## [2.3.0] - 2023-12-06
 
--- a/cls/SourceControl/Git/Change.cls
+++ b/cls/SourceControl/Git/Change.cls
@@ -78,8 +78,15 @@ ClassMethod IsUncommitted(Filename, ByRef ID) As %Boolean
 }
 
 /// Goes through Uncommitted queue and removes any items of action 'edit' or 'add' which are ReadOnly or non-existent on the filesystem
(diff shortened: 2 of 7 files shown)


isc-pbarton:
Something I found testing this on Unix with an old IRIS version (2019.1). Setting up git-source-control with Configure() fails:
`ERROR #5002: ObjectScript error: <METHOD DOES NOT EXIST>zOnAfterConfigure+39^SourceControl.Git.Settings.1 *Sync,%SYSTEM.WorkMgr`
I think this is a regression caused by #255. The Sync method was added in IRIS 2021.2, which is after the /ENV flag was added to $zf(-100).

isc-tleavitt:
@isc-pbarton sorry for the slow response - if you can look into that and tack on a fix it'd be great

isc-pbarton (approved):
I tacked on a fix to the unrelated issue on older IRIS versions, and smoke-tested on Unix with IRIS 2019.1 and Windows with IRIS 2023.2

Issue #300, opened 2023-12-07
PullEventHandler returns a Success Status even if an error was thrown during the process


isc-adepedro:
SourceControl.Git.API:Pull output:
```
w ##class(SourceControl.Git.API).Pull()

Pulling from branch: test
Fetch done
Files that will be modified by git pull: 
    M.gitlab-ci.yml
    DHS.Local.Hospital.GetPatientInfoProcess.CLSsrc/HS/Local/Hospital/GetPatientInfoProcess.cls
Updating 409377f..48f9b3e
                         Fast-forward
                                      .gitlab-ci.yml                                  | 140 ++++++++++++------------
                                                                                                                     src/HS/Local/Hospital/GetPatientInfoProcess.cls |  48 --------
                                                           2 files changed, 70 insertions(+), 118 deletions(-)
                                                                                                               delete mode 100644 src/HS/Local/Hospital/GetPatientInfoProcess.cls


.gitlab-ci.yml was not imported into the database and will not be compiled. 
ERROR importingHS.Local.Hospital.GetPatientInfoProcess.CLS

ERROR #5012: File '/home/irisowner/confstream/hospital/src/HS/Local/Hospital/GetPatientInfoProcess.cls' does not exist
Compilation started on 12/07/2023 13:25:33 with qualifiers 'cukb'
Class HS.Local.Hospital.GetPatientInfoProcess is up-to-date.
Compilation finished successfully in 0.050s.
1
```

The return of the previous execution is **1**

isc-tleavitt:
Might make sense to handle via the same PR as #299

isc-tleavitt:
#320 fixes this.

PR #320, opened 2024-01-24
Handle file deletion in incremental load PullEventHandler and throw errors (if any) instead of returning Success Status

### LINKED ISSUE #299: Incremental load PullEventHandler needs to handle deletes
Steps to reproduce:
* Delete a class
* w ##class(SourceControl.Git.API).Pull()

The deletion is detected by Git/extension ("D" Prefix):
```
Pulling from branch: test
Fetch done
Files that will be modified by git pull: 
    M.gitlab-ci.yml
    DHS.Local.Hospital.GetPatientInfoProcess.CLSsrc/HS/Local/Hospital/GetPatientInfoProcess.cls
Updating 409377f..48f9b3e
                         Fast-forward
                                      .gitlab-ci.yml                                  | 140 ++++++++++++------------
                                                                                                                     src/HS/Local/Hospital/GetPatientInfoProcess.cls |  48 --------
                                                           2 files changed, 70 insertions(+), 118 deletions(-)
                                                                                                               delete mode 100644 src/HS/Local/Hospital/GetPatientInfoProcess.cls


.gitlab-ci.yml was not imported into the database and will not be compiled.
```
But an error is thrown right after that:
`ERROR #5012: File '/home/irisowner/confstream/hospital/src/HS/Local/Hospital/GetPatientInfoProcess.cls' does not exist`

isc-vgao:
This fixes #299 , #300 
Testing Steps:
1. Delete a class in a repo.
2. Run
```
w ##class(SourceControl.Git.API).Pull()
```
3. Verify that no error messages (Error # ...) are printed in terminal. Instead, "<name of file deleted in step 1> was not imported into the database and will not be compiled." should be printed in terminal.

### DIFF
--- a/cls/SourceControl/Git/PullEventHandler/IncrementalLoad.cls
+++ b/cls/SourceControl/Git/PullEventHandler/IncrementalLoad.cls
@@ -17,8 +17,12 @@ Method OnPull() As %Status
         if ((internalName = "") && (..ModifiedFiles(i).changeType '= "D")) {
             write !, ..ModifiedFiles(i).externalName, " was not imported into the database and will not be compiled. "
         } elseif (..ModifiedFiles(i).changeType = "D") {
-            $$$ThrowOnError(..DeleteFile(internalName))
-            write !, ..ModifiedFiles(i).externalName, " was deleted."
+            set sc = ..DeleteFile(internalName)
+            if sc {
+                write !, ..ModifiedFiles(i).externalName, " was deleted."
+            } else {
+                write !, "Deletion of ", ..ModifiedFiles(i).externalName, " failed."
+            }
         } else {
             set compilelist(internalName) = ""
             set nFiles = nFiles + 1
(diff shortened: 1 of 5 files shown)


cls/SourceControl/Git/PullEventHandler/IncrementalLoad.cls
```diff
@@ -14,12 +14,13 @@ Method OnPull() As %Status
 
     for i=1:1:$get(..ModifiedFiles){
         set internalName = ..ModifiedFiles(i).internalName
-        if ((internalName = "") && (..ModifiedFiles(i).changeType '= "D")) {
+        if (((internalName = "") && (..ModifiedFiles(i).changeType '= "D")) || (..ModifiedFiles(i).changeType = "D")) {
```
isc-tleavitt, line 17: Tweak here: if changeType is D, we need to delete the item in the database (to *really* handle file deletion)

isc-tleavitt (changes requested):
Great simple solution - thank you! Just one minor change needed.

cls/SourceControl/Git/PullEventHandler/IncrementalLoad.cls
isc-tleavitt: Note: this entails calling $System.OBJ.Delete for a class, $System.OBJ.DeletePage for a CSP file, and %Library.RoutineMgr:Delete otherwise. Would probably be worth adding a helper method for that, if we don't already have one.

isc-tleavitt (changes requested):
One final note on error handling: as implemented we ignore whether the item was deleted or not, and then always say it was deleted. We should instead attempt to delete the item, and warn (without throwing an exception) if the delete fails (most likely because the item had already been deleted / didn't exist). So DeleteFile should return the status (possibly an error) from the Delete method used, but IncrementalLoad shouldn't $$$ThrowOnError - should just print a warning instead of "was deleted" if the status was an error.

isc-tleavitt (approved):

PR #324, opened 2024-03-01
Git pull correctly actions modified files

### LINKED ISSUE #323: Pull treats added files from remote branch as deletes
Caused by #320 

To replicate:

- add a file on one instance, commit, and push to a remote
- on a separate instance with the same remote, use the "Pull changes from remote branch" option in Studio/VSCode
- from the output, I see it is trying to delete the new file instead of load it

isc-pbarton:
Fixes #323 

The git diff now shows differences from the opposite side. So a file added on the remote branch is treated as an add rather than a delete, and vice versa.
I think no changelog required because this fixes unintended behavior from an unreleased change #320.


isc-tleavitt (approved):
Makes sense. Thank you!

PR #330, opened 2024-04-19
Documentation for offline installation

### LINKED ISSUE #310: Document and better support installation without network access to OEX registry
Per thread under https://community.intersystems.com/post/git-shared-development-environments#comment-247086

isc-pbarton:
Fixes #310

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,11 @@
     ```
     zpm "install git-source-control"
     ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,6 +14,11 @@
     ```
     zpm "install git-source-control"
     ```
+    To install on an environment without access to the internet, download the tar.gz file from the [releases](https://github.com/intersystems/git-source-control/releases) page. Copy the archive onto a file system the IRIS instance has access to and extract it. Use the package manager to load the release from that directory.
+    ```
+    tar -xf /path/to/archive/git-source-control-release.tar.gz
```
isc-tleavitt, line 19: Do you actually need to do this? I think zpm can load a .tar.gz directly.

isc-tleavitt (approved):

PR #334, opened 2024-04-30
Get ready for IPM v0.9.0

isc-tleavitt:
Class names have changed and git-source-control goes under the hood in IPM, so it needs to be adjusted.


isc-tleavitt:
@isc-pbarton can you please review this?

isc-pbarton (approved):

PR #345, opened 2024-05-31
Fixes #189 Export and Add ignore mapped

### LINKED ISSUE #189: Make sure "Export All" can/does (by default) ignore mapped items
We need a new setting - on by default - to treat mapped items as read-only. If set, this setting should also prevent mapped items from being added to source control, and should cause the extension to skip over mapped items in "Export All"

From discussion with @mdavidovichHst

isc-hwojnick:
Mapped files are now ignored when adding or exporting if the read-only mapped setting is enabled (on by default)

### DIFF
--- a/csp/gitprojectsettings.csp
+++ b/csp/gitprojectsettings.csp
@@ -47,6 +47,10 @@ body {
     border-left: 0px;
 }
 
+.custom-switch-no-border {
+    border: 0px;
+}
+
 .neutral-feedback {
     width: 100%;
     margin-top: 0.25rem;
@@ -70,6 +74,13 @@ body {
         for param="gitBinPath","namespaceTemp","privateKeyFile","pullEventClass","percentClassReplace" {
             set $Property(settings,param) = $Get(%request.Data(param,1))
         }
+
+        if ($Get(%request.Data("mappedItemsReadOnly", 1)) = 1) {
+            set settings.mappedItemsReadOnly = 1
+        } else {
+            set settings.mappedItemsReadOnly = 0
+        }
+        set ^mtemphw($i(^mtemphw)) = $get(%request.Data("mappedItemsReadOnly", 1))
         set i = 1
         set param = "NoFolders"
         kill settings.Mappings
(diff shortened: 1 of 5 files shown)


csp/gitprojectsettings.csp
```diff
@@ -70,6 +74,13 @@ body {
         for param="gitBinPath","namespaceTemp","privateKeyFile","pullEventClass","percentClassReplace" {
             set $Property(settings,param) = $Get(%request.Data(param,1))
         }
+
+        if ($Get(%request.Data("mappedItemsReadOnly", 1)) = 1) {
+            set settings.mappedItemsReadOnly = 1
+        } else {
+            set settings.mappedItemsReadOnly = 0
+        }
+        set ^mtemphw($i(^mtemphw)) = $get(%request.Data("mappedItemsReadOnly", 1))
```
isc-tleavitt, line 83: Debug code - please remove.

csp/gitprojectsettings.csp
```diff
@@ -93,8 +104,7 @@ body {
         <input type="hidden" name="gitsettings" value="1" />
         <div class="col-sm-12"><br></div>
         <div class="row">
-            <div class="col-sm-1"></div>
-            <div class="col-sm-8">
+            <div class="offset-sm-1 col-sm-8">
```
isc-tleavitt, line 107: Thanks for fixing these!

csp/gitprojectsettings.csp
```diff
@@ -255,16 +258,36 @@ body {
...
+            <label for="percentClassReplace" class="offset-sm-1 col-sm-3 col-form-label"  data-toggle="tooltip" data-placement="top" title="Character(s) to replace '%' with for percent classes while exporting them out to the filesystem. By default, the '%' is removed.">'%' Replacement on Export</label>
             <div class="col-sm-7">
                 <input type="text" class="form-control" id="percentClassReplace" name="percentClassReplace" value='#(..EscapeHTML(settings.percentClassReplace))#' placeholder="_, __, <empty>, etc."/>
             </div>
         </div>
+        
+        
+        <div class="form-group row mb-3">
+            <label for="mappedItemsReadOnly" class="offset-sm-1 col-sm-3 col-form-label" data-toggle="tooltip" data-placement="top" title="Whether items mapped from another database should be read-only. If enabled, mapped items won't be saved to source control or exported. NOTE: These are different from the mappings configured in this settings page"> Treat Mapped Items as Read-only</label>
```
isc-tleavitt, line 269: "mapped from another database" could be more precisely phrased as "mapped from a database other than this namespace's default routine database"

isc-tleavitt (changes requested):
In general, for each non-documentation PR, please also include an update to CHANGELOG.md following the format you see there.

isc-tleavitt (approved):

Issue #317, opened 2024-01-12
Can't use with IPM in a multi-developer setting


isc-tleavitt:
Ran into this accidentally today:
Suppose you're working in package A, then switch context to Package B. RefreshUncommitted() in a package B context will clear out uncommitted changes from Package A. It shouldn't touch them. (Simple enough - only clear out things matching the root.) There may be some downstream side effects of this in e.g. the git UI.

Nobody's complained about this yet, probably because users are either working with IPM locally *or* are on a multi-developer environment not using IPM. Still worth fixing.

PR #347, opened 2024-06-04
Issue #317 IPM Multi Developer Setting Issue Resolved

isc-hwojnick:
Files that belong to other Git enabled IPM packages are now skipped when removing files in ##class(SourceControl.Git.Change).RefreshUncommitted().

### DIFF
--- a/cls/SourceControl/Git/Change.cls
+++ b/cls/SourceControl/Git/Change.cls
@@ -128,12 +128,16 @@ ClassMethod RefreshUncommitted(Display = 0, IncludeRevert = 0, Output gitFiles, 
     // Remove entries in the uncommitted queue that don't correspond to changes as tracked by git
     set filename="", filename=$order(tFileList(filename),1,action)
     while (filename'="") {       
-        Set ^mtemphw("filename", $i(^mtemphw("filename"))) = filename
         set examine=$select(action="add":1,action="edit":1,action="delete":1, IncludeRevert&&(action="revert"):1,1:0)
         if 'examine set filename=$order(tFileList(filename),1,action) continue
 
+        set context = ##class(SourceControl.Git.PackageManagerContext).%Get()
+        set packageRoot = context.Package.Root
         set InternalName = ##class(SourceControl.Git.Utils).NameToInternalName(filename,0,0)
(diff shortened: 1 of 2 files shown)


cls/SourceControl/Git/Change.cls
```diff
@@ -131,8 +131,14 @@ ClassMethod RefreshUncommitted(Display = 0, IncludeRevert = 0, Output gitFiles,
         set examine=$select(action="add":1,action="edit":1,action="delete":1, IncludeRevert&&(action="revert"):1,1:0)
         if 'examine set filename=$order(tFileList(filename),1,action) continue
 
+        set context = ##class(SourceControl.Git.PackageManagerContext).%Get()
+        set packageRoot = context.Package.Root
         set InternalName = ##class(SourceControl.Git.Utils).NameToInternalName(filename,0,0)
 
+        // skip files belonging to other git enabled packages
+        if ($EXTRACT(filename, 1, $LENGTH(packageRoot)) = packageRoot) continue
```
isc-tleavitt, line 139: I think this is wrong. Should get context and packageRoot outside of the loop, and continue if one of the following is true:
* context.Package is an object and filename is not under the package root
* context.Package is not an object and filename is not under the namespace-default root

Note, this might have side effects in the WebUI too - you'll need to make sure that the endpoint that lists uncommitted items properly filters to the current package context too. (Just something to test: if you have uncommitted changes from multiple package contexts, what shows in the web UI in each of those contexts?)

isc-tleavitt (changes requested):
I don't think this does what you intend it to.

cls/SourceControl/Git/Change.cls
isc-tleavitt: Actually, `##class(SourceControl.Git.Utils).TempFolder()` already covers all this:

```
ClassMethod TempFolder() As %String
{
    set context = ##class(SourceControl.Git.PackageManagerContext).%Get()
    if context.IsInGitEnabledPackage {
        quit context.Package.Root
    }
    quit ..DefaultTempFolder()
}
```

isc-tleavitt (approved):

Issue #351, opened 2024-06-06
Add branch name to SourceControl.Git.PullEventHandler instance


isc-tleavitt:
A new property, Branch, should be added to SourceControl.Git.PullEventHandler and be set to the name of the current checked-out branch prior to OnPull() being called.

PR #353, opened 2024-06-06
Issue #351 added branch property to the PullEventHandler class

isc-hwojnick:
The new branch property in the PullEventHandler class now defaults to holding the value of the current branch before OnPull() is called.


isc-tleavitt (approved):
Need to resolve CHANGELOG.md, otherwise good to go.

Issue #349, opened 2024-06-06
Add configurable "Basic" and "Expert" modes


isc-tleavitt:
From a few recent internal discussions: many users, particularly in a healthcare interoperability setting, git is very foreign. It would be better if the git integration could be made more seamless/automatic.

As our motivating use case, we have a single-developer, mostly interoperability-driven development environment. Merges in any direction happen through a web-based UI for the remote which allows for conflict resolution. There is a CI/CD pipeline that pushes changes to developers' environments automatically.

To support this use case, we should have an option to choose between "Basic" and "Expert" modes. This mode is a user-level setting configurable via the settings page, with a system-wide default that applies if it is not configured for the current user. "Expert" mode matches all of the current functionality. In "Basic" mode:
* git add of newly-created items is automatic (this could also be made a system-level setting)
* "Push to remote branch", "Fetch from remote", "Pull changes from remote branch" are boiled down to a single "Sync" menu item which fetches, pulls, commits *all* of the user's uncommitted changes (with an optional message defaulting to something reasonable), then pushes.
* "Create new branch", "Check out an existing branch", and the export/import options are removed from the menu.

PR #355, opened 2024-06-10
Issue #349 added basic and expert modes to the application

isc-hwojnick:
A basic mode was added to the application so that users less familiar with Git can have simplified workflows


csp/gitprojectsettings.csp
```diff
@@ -403,6 +435,21 @@ body {
             </div>
         </div>
 
+        <div class="form-group row mb-3">
+            <label for="basicMode" class="offset-sm-1 col-sm-3 col-form-label" data-toggle="tooltip" data-placement="top" title="Enable basic mode, greatly simplifying the Git Source Control package, requiring no knowledge of git">Basic Mode</label>
```
isc-tleavitt, line 439: Reword:
Enable basic mode, which reduces the number of menu options and requires less knowledge of git. This is suitable for a single-developer (non-shared) namespace.

csp/gitprojectsettings.csp
```diff
@@ -263,6 +275,26 @@ body {
             </div>
         </div>
         
+        <div class="form-group row mb-3">
+            <label for="systemBasicMode" class="offset-sm-1 col-sm-3 col-form-label" data-toggle="tooltip" data-placement="top" title="Whether basic mode should be applied as the system-wide default"> Basic Mode as System Default</label>
```
isc-tleavitt, line 279: Reword:
Enable basic mode for all users, which reduces the number of menu options and requires less knowledge of git. This is suitable for a single-developer (non-shared) namespace.

csp/gitprojectsettings.csp
isc-tleavitt: We need to think about the interaction of this with system-wide. This is really tri-state and should be represented via a radio button set: enabled, disabled, default (follows system mode and corresponds to unset global) - this will need changes in the settings methods too.

cls/SourceControl/Git/Utils.cls
```diff
@@ -1473,7 +1510,7 @@ ClassMethod RunGitCommand(command As %String, Output errStream, Output outStream
 ClassMethod RunGitCommandWithInput(command As %String, inFile As %String = "", Output errStream, Output outStream, args...) As %Integer
 {
     // Special case: git --version is used internally even when the settings incorporated here may be invalid/unspecified.
-    if (command '= "--version") {
+    if (command '= "--version") && (command '= "remote") {
```
isc-tleavitt, line 1513: I'm concerned with omitting remote here - this should follow IPM context and therefore depends on the root. Can you explain?

cls/SourceControl/Git/Extension.cls
```diff
@@ -236,15 +254,17 @@ Method OnAfterSave(InternalName As %String, Object As %RegisteredObject = {$$$NU
...
-            $$$ThrowOnError(##class(Utils).ExportItem(InternalName))
+        if ##class(SourceControl.Git.Utils).IsNamespaceInGit() && ..IsInSourceControl(InternalName) {
+            set filename = ##class(SourceControl.Git.Utils).FullExternalName(InternalName)
+            $$$ThrowOnError(##class(SourceControl.Git.Utils).RemoveRoutineTSH(InternalName))
+            $$$ThrowOnError(##class(SourceControl.Git.Utils).ExportItem(InternalName))
             if '##class(SourceControl.Git.Change).IsUncommitted(filename) {
                 $$$ThrowOnError(##class(SourceControl.Git.Change).SetUncommitted(filename, "edit", InternalName, $username, "", 1, "", "", 0))
             }
+        } elseif ##class(SourceControl.Git.Utils).BasicMode() {
+            do ..AddToSourceControl(InternalName)
```
isc-tleavitt, line 267: I think this belongs in UserAction/AfterUserAction for creation of new things (those are types of system events) rather than OnAfterSave - if a user elects to remove (not track) a certain file we should reflect that.

cls/SourceControl/Git/Utils.cls
```diff
@@ -322,6 +334,22 @@ ClassMethod SwitchBranch(targetBranchName As %String) As %Status
     quit $$$OK
 }
 
+ClassMethod Sync() As %Status
+{
+    write "Syncing local repository..."
+    if '..HasRemoteRepo() {
+        write "No remote repository configured: skipping fetch, pull and push"
+        do ..Commit("", "Committing all files through sync command")
```
isc-tleavitt, line 342: This should:
* Build a list of files with tracked changes (staged/unstaged, we want all of them)
* Build a reasonable commit message based on that
* Offer the user the opportunity to customize the commit message (via a different "Action")

The flow for this is going to be building out the list, commit message, offering it to the user in UserAction and then actually doing the Fetch/Pull/Commit/Push in AfterUserAction (from the extension itself - might still want to have helper methods in Utils for consistency, e.g. PreSync and DoSync)

isc-tleavitt (changes requested):
@isc-hwojnick a few points of feedback here - I'm impressed with how quickly you turned around this much!

cls/SourceControl/Git/Extension.cls
isc-hwojnick: Right. You make a good point here! Not sure what I was thinking at the time.

cls/SourceControl/Git/Extension.cls
```diff
@@ -49,15 +50,20 @@ Method UserAction(Type As %Integer, Name As %String, InternalName As %String, Se
         set Name = "%SourceMenu,Status"
     } 
 
+    if Type = 1, Name = 1 {
+        do ..AddToSourceControl(InternalName)
```
isc-tleavitt, line 54: Initial thought: We only want to do this automatically in basic mode
On further review: Actually we should just always do this automatically, you're right. I think you might also need to pick up Name = 7 though.

cls/SourceControl/Git/Extension.cls
```diff
@@ -49,15 +50,20 @@ Method UserAction(Type As %Integer, Name As %String, InternalName As %String, Se
         set Name = "%SourceMenu,Status"
     } 
 
+    if Type = 1, Name = 1 {
```
isc-tleavitt, line 53: "," is not modern ObjectScript syntax - please change to:
if (Type = 1) && ((Name = 1) || (Name = 7))

git-webui/release/share/git-webui/webui/js/git-webui.js
```diff
@@ -2147,7 +2147,9 @@ webui.ChangedFilesView = function(workspaceView, type, label) {
         var col = type == "working-copy" ? 1 : 0;
         webui.git("status -u --porcelain", function(data) {
             $.get("api/uncommitted", function (uncommitted) {
+                console.log(uncommitted);
```
isc-tleavitt, line 2150: Please remove console.log, this will break in Studio.

git-webui/release/share/git-webui/webui/js/git-webui.js
```diff
@@ -2147,7 +2147,9 @@ webui.ChangedFilesView = function(workspaceView, type, label) {
...
                 var uncommittedItems = JSON.parse(uncommitted)["current user's changes"];
+                console.log(uncommittedItems);
```
isc-tleavitt, line 2152: Remove here too

git-webui/src/share/git-webui/webui/js/git-webui.js
```diff
@@ -2147,7 +2147,9 @@ webui.ChangedFilesView = function(workspaceView, type, label) {
         var col = type == "working-copy" ? 1 : 0;
         webui.git("status -u --porcelain", function(data) {
             $.get("api/uncommitted", function (uncommitted) {
+                console.log(uncommitted);
```
isc-tleavitt, line 2150: ... and here

git-webui/src/share/git-webui/webui/js/git-webui.js
```diff
@@ -2147,7 +2147,9 @@ webui.ChangedFilesView = function(workspaceView, type, label) {
...
                 var uncommittedItems = JSON.parse(uncommitted)["current user's changes"];
+                console.log(uncommittedItems);
```
isc-tleavitt, line 2152: ... and here

cls/SourceControl/Git/Utils.cls
```diff
@@ -1533,6 +1630,29 @@ ClassMethod RunGitCommandWithInput(command As %String, inFile As %String = "", O
...
+            set commitMsg = commitMsg _ "added " _ file _ ", "
+        } elseif action = "M" {
+            set commitMsg = commitMsg _ "modified " _ file _ ", "
+        } elseif action = "D" {
+            set commitMsg = commitMsg _ "deleted " _ file _ ", "
+        } elseif action = "R" {
+            set commitMsg = commitMsg _ "renamed " _ file _ ", "
+        }
+    }
+    quit commitMsg
```
isc-tleavitt, line 1653: Useful pattern for things like this to avoid the trailing ", " that you'll currently have:
set commitMsg = ""

In loop:
set commitMsg = commitMsg_$ListBuild(moreText)

At end:
quit $ListToString(commitMsg,", ")

cls/SourceControl/Git/Utils.cls
```diff
@@ -1533,6 +1630,29 @@ ClassMethod RunGitCommandWithInput(command As %String, inFile As %String = "", O
...
+{
+    set commitMsg = ""
+
+    set iterator = filesWithActions.%GetIterator()
+    while iterator.%GetNext(,.value,) {
+        set action = value.%Get("action")
+        set file = value.%Get("file")
+
+
+        if action = "A" {
```
isc-tleavitt, line 1643: Perhaps restructure as:

```
set oneFileMsg = $Case(action,
  "A":"added",
  "M":"modified",
  "D":"deleted",
  "R":"renamed",
  :"???")_" "_file
```

It's a bit less code and will also handle (to some extent) other operations, if they happen to exist - do you have an exhaustive list of codes?

cls/SourceControl/Git/WebUIDriver.cls
```diff
@@ -200,6 +200,42 @@ ClassMethod Uncommitted() As %SystemBase
...
+            if ($$$ISOK(sc)) && ($data(tAction)&&(UncommittedUser=$username)) {
+                set edit = {}
+                do edit.%Set("file", $listget(fileData, 2))
+            do edit.%Set("action", $listget(fileData, 1))
+            do editedByCurrentUser.%Push(edit)
+            }
+        }
+    }
+
     do fileToOtherDevelopers.%Set("current user's changes", editedByCurrentUser)
```
isc-tleavitt, line 239: Since we're adding a new API, can we pick a more reasonable key here? e.g., call it "user", "others" rather than having long magic strings? Or better yet, since we have attributes already, do edit.%Set("user",UncommittedUser) and do edit.%Set("mine",UncommittedUser=$username)?

cls/SourceControl/Git/Settings.cls
```diff
@@ -32,6 +32,15 @@ Property gitUserEmail As %String(MAXLEN = 255) [ InitialExpression = {##class(So
 /// Attribution: Whether mapped items should be read-only, preventing them from being added to source control
 Property mappedItemsReadOnly As %Boolean [ InitialExpression = {##class(SourceControl.Git.Utils).MappedItemsReadOnly()} ];
 
+/// Attribution: Whether basic mode should be enabled for user ${username}, greatly simplifying the functionality of the package, requiring no knowledge of git
```
isc-tleavitt, line 35: Nit: this is not Attribution-related

cls/SourceControl/Git/Utils.cls
isc-tleavitt: @isc-hwojnick  I'm still concerned about this.

cls/SourceControl/Git/Extension.cls
isc-tleavitt: (and feel free to fix this anywhere you find it in old code!)

isc-tleavitt (changes requested):
@isc-hwojnick a few more changes needed here - we're almost totally there functionally, just need some code-level cleanup.

cls/SourceControl/Git/Utils.cls
isc-hwojnick: @isc-tleavitt I'm pretty sure I resolved this as you requested. Not sure why github doesn't show that

cls/SourceControl/Git/Settings.cls
isc-hwojnick: Oops was just trying to follow the pattern of comments

isc-pbarton, line 35:
```suggestion
/// Whether basic mode should be enabled for user ${username}, greatly simplifying the functionality of the package, requiring no knowledge of git
```

cls/SourceControl/Git/Settings.cls
```diff
@@ -32,6 +32,15 @@ Property gitUserEmail As %String(MAXLEN = 255) [ InitialExpression = {##class(So
...
+Property basicMode As %String [ InitialExpression = {##class(SourceControl.Git.Utils).BasicMode()} ];
+
+/// This does not map to a setting on the screen but is used to select the value to be displayed for the user's basic mode setting
```
isc-pbarton, line 38: The comment displays in the terminal prompt when running through Configure(), so this is currently somewhat confusing for the user:
```
Attribution: Whether basic mode should be enabled for user 'pbarton', greatly simplifying the functionality of the package, requiring no knowledge of git:
  0 => 0
This does not map to a setting on the screen but is used to select the value to be displayed for the user's basic mode setting:
  system => idk
The system's default mode. If true, the system defaults to basic mode:
  0 => 1
Settings saved.
```
I personally would remove the user-level setting for basic mode and just have the one system-level setting for basic mode. But I will defer to Tim if you have discussed otherwise.

cls/SourceControl/Git/Utils.cls
```diff
@@ -322,6 +347,69 @@ ClassMethod SwitchBranch(targetBranchName As %String) As %Status
     quit $$$OK
 }
 
+ClassMethod PreSync() As %String
+{
+    set uncommittedFilesWithAction = ##class(SourceControl.Git.WebUIDriver).UncommittedWithAction().%Get("user")
```
isc-pbarton, line 352: Nitpick: I would move this UncommittedWithAction() method into the Utils class because it's less tidy to have Utils referencing the WebUI code.

cls/SourceControl/Git/Utils.cls
```diff
@@ -322,6 +347,69 @@ ClassMethod SwitchBranch(targetBranchName As %String) As %Status
...
+    quit ..GenerateCommitMessageFromFiles(uncommittedFilesWithAction)
+}
+
+/// Commits all the files as needed by the Sync operation 
+ClassMethod SyncCommit(Msg As %String) As %Status
+{
```
isc-pbarton, line 358: I think the Sync option should also go through each item modified by the current user and stage it as well.

isc-pbarton (changes requested):

cls/SourceControl/Git/Utils.cls
isc-hwojnick: @isc-pbarton I added code to stage the files upon calling sync

isc-pbarton (approved):

isc-tleavitt:
@isc-hwojnick you should resolve conflicts locally and re-push.

isc-tleavitt (approved):

isc-tleavitt:
@isc-hwojnick I think this is good! Just resolve the conflicts locally and push back up, then we'll merge.

PR #368, opened 2024-06-20
Compile items after Import All

### LINKED ISSUE #362: Import All doesn't compile imported items
The "Import All" and "Import All (Force)" options do not compile a class/routine after importing it. So for example:
- I merge a remote branch containing a change to an interoperability production using Git WebUI
- I choose "Import All" expecting to see the new change
- The production config page still shows the old version since the class never got compiled.

Could go either way here, but I don't see much use for importing the item without compiling it.

isc-pbarton:
Fixes #362

### DIFF
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -1187,7 +1187,7 @@ ClassMethod ImportItem(InternalName As %String, force As %Boolean = 0, verbose A
             if ($extract(InternalName, 1) = "/"){
                 set sc = ..ImportCSPFile(InternalName)
             } else{
-                set sc = $system.OBJ.Load(filename,"-l-d")
+                set sc = $system.OBJ.Load(filename,"ck")
             }
         }
         if sc {
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -1187,7 +1187,7 @@ ClassMethod ImportItem(InternalName As %String, force As %Boolean = 0, verbose A
             if ($extract(InternalName, 1) = "/"){
                 set sc = ..ImportCSPFile(InternalName)
             } else{
-                set sc = $system.OBJ.Load(filename,"ck")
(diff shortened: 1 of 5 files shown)


raymond-rebbeck:
Please make this configurable. The codebase I work with requires some classes to be compiled in a particular order on first import for reasons that are not practical to address in the short term. I am concerned that this will cause problems for us.

isc-tleavitt:
@isc-pbarton the point @raymond-rebbeck raises is fair - let's make this configurable, but we want to default to compiling (particularly for the sake of less-technical interop users who wouldn't necessarily understand the implications).

cls/SourceControl/Git/Utils.cls
```diff
@@ -1230,7 +1230,7 @@ ClassMethod ImportItem(InternalName As %String, force As %Boolean = 0, verbose A
             if ($extract(InternalName, 1) = "/"){
                 set sc = ..ImportCSPFile(InternalName)
             } else{
-                set sc = $system.OBJ.Load(filename,"-l-d")
```
isc-tleavitt, line 1233: Interesting that we were using -l - I wonder if there was a reason for this. I agree that displaying output is reasonable, for sure.

Question - what's the behavior if a single item fails to load/compile?

A better architecture here would be to maintain a list of things that were loaded by specific calls to ImportItem and then compile all of them at the same time after all have been loaded. As written this is likely to hit dependencies (especially on initial load); with compilation at the end it at least *should* be possible for a developer to fix compilation dependencies with the right values for System, CompileAfter, DependsOn, etc.

isc-tleavitt (changes requested):
Two points here: configurability is needed (defaulting to compile but should be able to turn it off), and we should compile everything after loading everything rather than compiling each thing as it's loaded.

cls/SourceControl/Git/Utils.cls
isc-tleavitt: Note from discussion today: just use PullEventHandler

isc-tleavitt (approved):
Looks good

PR #376, opened 2024-06-24
Fixes #366 Add merge to sync flow

### LINKED ISSUE #366: Basic mode option to rebase from remote branch
Use case is in trunk-based development, where a developer is working in a feature branch, but wants to regularly merge in changes from a different remote branch (e.g. production).

In Basic mode, add a step to the sync option: it should fetch, pull, merge from a remote branch, commit, then push. There should be a new setting that specifies which remote branch should be merged from.

isc-hwojnick:
Now allows users to configure a default branch to merge from during the sync flow before pushing

### DIFF
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -420,8 +420,9 @@ ClassMethod Sync(Msg As %String) As %Status
         do ..Fetch()
         do ..Pull()
         do ..SyncCommit(Msg)
-        do ..MergeDefaultRemoteBranch()
         do ..Push()
+        do ..MergeDefaultRemoteBranch()
+        
     }
     
     quit $$$OK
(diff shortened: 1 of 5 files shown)


cls/SourceControl/Git/Utils.cls
```diff
@@ -404,6 +420,7 @@ ClassMethod Sync(Msg As %String) As %Status
         do ..Fetch()
         do ..Pull()
         do ..SyncCommit(Msg)
+        do ..MergeDefaultRemoteBranch()
```
isc-tleavitt, line 423: I think we should Push before attempting to merge (as well as after). That way if there's a conflict it can be resolved in the remote with the latest from the feature branch already available there. A subsequent Pull after such a resolution should be able to get things back to a consistent state.

isc-tleavitt (changes requested):

isc-tleavitt (changes requested):
One last thing here, then I think we're good.

isc-tleavitt (approved):

PR #381, opened 2024-06-25
ignore deletions from merge, rebase, etc.

isc-pbarton:
Currently we sync changes to IRIS from a merge, rebase, etc. by running diff beforehand to find the modified items. This doesn't work well if a file was added to the local branch but not pushed to the remote branch - that add is treated as a delete by the diff, which causes the item to be deleted from IRIS.
For now, we are just ignoring deletions.
In the long term, instead of using "diff" to determine modified items we will need to look at the output of merge, rebase, etc. to determine when things are actually being deleted and added.


isc-tleavitt:
This is supplanted by #384

PR #384, opened 2024-06-26
Merge, Pull and Rebase now don't use diff to synchronize file system with IRIS

isc-hwojnick:
Merge, Pull and Rebase will now use the output of the command itself to drive their synchronization with IRIS. 

ex:
the merge command will output:
> Merge made by the 'ort' strategy. 
> cls/package/newb4.cls | 5 ----- 
> cls/package/newb5.cls | 5 ----- 
> cls/package/newb7.cls | 5 +++++ 
> 3 files changed, 5 insertions(+), 10 deletions(-)
> delete mode 100644 cls/package/newb4.cls 
> delete mode 100644 cls/package/newb5.cls 
> create mode 100644 cls/package/newb7.cls 

In this example, the last 3 lines of the output are used to sync iris with the filesystem

### DIFF
--- a/cls/SourceControl/Git/Util/Buffer.cls
+++ b/cls/SourceControl/Git/Util/Buffer.cls
@@ -281,3 +281,4 @@ Method UsePreviousDeviceAndSettings() [ Internal, Private ]
 }
 
 }
+
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -1565,13 +1565,16 @@ ClassMethod RunGitCommandWithInput(command As %String, inFile As %String = "", O
 
     set newArgs($increment(newArgs)) = command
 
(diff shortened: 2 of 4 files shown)


isc-pbarton (changes requested):
Merge and checkout of local and remote branches are working great for me.
Pull is having the issue from before where deletes/adds on the local branch get undone - but that is preexisting behavior for as long as the pull function was added. We could create a new issue to address that later by using the same logic as merge and checkout.
The preview pull on pull.csp is broken again and doesn't display any output. That should be fixed, I think by printing the output streams after preview in %zpkg.isc.sc.git.Socket

isc-hwojnick:
As for the previewing, I think I can fix that

isc-pbarton (approved):

isc-tleavitt (approved):

PR #390, opened 2024-06-28
Fix production conflicts automatically (and more sync tweaks)

### LINKED ISSUE #371: Sync option should not prompt for commit message if no changes
Say git-source-control is configured in Basic mode and I have no uncommitted changes. If I use the Sync command, I will be prompted for a commit message. It will then fetch, pull, see I have no changes to commit, and finally push.

If possible let's skip that commit message prompt if there are no changes to commit.

### LINKED ISSUE #380: Error running "Import"
image

### LINKED ISSUE #387: Discard changes should compile
Internal reference: see comments on DP-424937

### LINKED ISSUE #389: Git diff-based loading should do the right thing with detected renames
Git diff might detect renames. These look like (e.g., from https://stackoverflow.com/questions/40702017/what-does-the-number-in-rnumber-after-a-git-rename-mean):

R079    Web/Views/Shared/Blocks/ChristmasLotteryBlock.cshtml    Web/Views/Shared/Blocks/ChristmasLotteryBlock/Index.cshtml
R076    Web/Static/blocks/_block_christmas-lottery-intro.html   Web/Views/Shared/Blocks/ChristmasLotteryBlock/_intro.cshtml
R099    Web/Static/blocks/_block_christmas-lottery-popup.html   Web/Views/Shared/Blocks/ChristmasLotteryBlock/_popup.cshtml
M       Web/Web.csproj

For purposes of IRIS, the file on the left is a delete and the file on the right is an add. Right now our diff-based incremental load tries to act on the file on the left, which doesn't exist, which can lead to very odd inexplicable failures after e.g. changing branches from the WebUI.

isc-tleavitt:
When we sync, we now automatically resolve differences in the production definition.
Also cleans up a few rough edges / edge cases with sync and import.

Fixes #380
Fixes #371
Fixes #389
Fixes #387

Pending tasks:
- [x] preserve commit messages in production merge
- [x] support resolution on sequential commits - skipping for now, won't happen with a single file
- [x] ensure a git rebase --continue that modifies the production reloads and compiles it
- [x] fix handling of rename on diff for sync


isc-tleavitt:
Hit one issue: a git rebase --continue may itself trigger a non-conflicting change to the production. We need to ensure it is reloaded and compiled in this case to avoid the really annoying error alert.

cls/SourceControl/Git/Util/ProductionConflictResolver.cls
```diff
@@ -0,0 +1,156 @@
...
+    }
+    If '..logStream.AtEnd {
+        $$$ThrowStatus($$$ERROR($$$GeneralError,"Multiple files had merge conflicts; cannot resolve intelligently."))
+    }
+    Set internalName = ##class(SourceControl.Git.Utils).NameToInternalName(..productionFile)
+    If ($Piece(internalName,".",*) '= "CLS") {
+        $$$ThrowStatus($$$ERROR($$$GeneralError,"File with conflict is not a class."))
+    }
+    Set ..productionClassname = $Piece(internalName,".",1,*-1)
+    If '$$$comClassDefined(..productionClassname) && '$ClassMethod(..productionClassname,"%Extends","Ens.Production") {
```
isc-pbarton, line 68: Should this be an or instead of and? Say class with merge conflict is not defined, will try to call %Extends and throw error.

isc-tleavitt: Good catch!

isc-pbarton (approved):

PR #394, opened 2024-06-28
Fix #371 - now doesn't prompt for commit msg if no uncomitted files

### LINKED ISSUE #371: Sync option should not prompt for commit message if no changes

isc-hwojnick:
Syncing only prompts users for a commit message if there are uncommitted files

Fixes #371


isc-tleavitt:
@isc-hwojnick this is the right general direction, but there's a bonus - if there's no commit message we have the opportunity for UserAction to alert the user on error. This is going to end up conflicting with my work in #390 so I'm going to pull these changes over there to finish and close out this PR.

isc-tleavitt:
Folded into #390

PR #396, opened 2024-07-02
Developer sourcecontrol select privileges

### LINKED ISSUE #365: Automatically give %Developer SELECT privileges on %Studio_SourceControl.Change and SourceControl_Git.Change
Might need to do this in a %z class with $$$AddAllRoleTemporary

This seems to be new to 2024.1 - need to understand why.

isc-hwojnick:
Fixes #365 

Now properly retrieves the list of uncommitted files made by another user


isc-tleavitt (approved):
This does the trick! (I tested in HCC to be sure.)

PR #410, opened 2024-07-08
update uncommitted now removes other user's uncommitted files if they're committed

### LINKED ISSUE #408: Some git operations aren't updating SourceControl_Git.Change
git commit doesn't update the SourceControl_Git.Change table. This leads to users being locked out from editing files another user worked on even after they're committed.

isc-hwojnick:
Fixes #408


isc-tleavitt (approved):
Seems like some minor pollutants in the diff but don't want to waste time on that. Looking good to me.

PR #416, opened 2024-07-10
Fix JavaScript errors in Studio

raymond-rebbeck:
Use 'substring' instead of 'startsWith' to prevent errors in Studio when performing actions such as refreshing remotes.

'startsWith' is not supported by Internet Explorer.


isc-tleavitt (approved):

isc-tleavitt:
@raymond-rebbeck thank you once again! We should have caught this before release.

PR #419, opened 2024-07-10
Fix bugs pertaining to adding files

### LINKED ISSUE #404: Add an item to source control does not update uncommitted queue
- Create a new item in Studio / VS Code
- Use the "Add" menu option to add it to source control
- In the Workspace view of the web UI, the change shows the icon saying it belongs to a different user.
- Use the "Status" menu option
- Now the Workspace view will show the change as belonging to me.

The cause is that AddToSourceControl does not add the item to the SourceControl.Git.Change table.

### LINKED ISSUE #407: New items should be automatically added to source control in Basic Mode
If I add a new item through Studio / VS Code / Interoperability portals / wherever, and I use the Sync option, that new item is not committed unless I explicitly add it to source control. In Basic mode, new items should get automatically added to source control.

isc-hwojnick:
Fixes #404
Fixes #407

Files are auto added to source control (and not staged) and show in the uncommitted queue. This ends up resolving #404 and #407 in one fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed JS errors in Studio on certain operations (#416)
+- Files are added to source control upon creation properly (#404)
+- Files show in uncommitted queue when automatically added (#407)
 
 ## [2.4.0] - 2024-07-08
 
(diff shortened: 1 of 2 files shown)


isc-tleavitt (changes requested):
That was easy! Just needs a CHANGELOG.md entry.

isc-tleavitt (approved):

PR #420, opened 2024-07-10
disallows the user from choosing the add menu option on an uncompiled file

### LINKED ISSUE #405: Adding uncompiled class to source control fails
Add a new class through Studio or VS Code. Without compiling it yet, choose the "Add" option in the Source Control menu. I get an error message: `ERROR #16005: Document 'pbarton.class.cls' does NOT exist`

isc-hwojnick:
Fixes #405

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed JS errors in Studio on certain operations (#416)
-- Add menu option disabled for uncompiled files (#420)
+- Add menu option disabled for unsaved files (#420)
 
 ## [2.4.0] - 2024-07-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed JS errors in Studio on certain operations (#416)
+- Add menu option disabled for uncompiled files (#420)
 
 ## [2.4.0] - 2024-07-08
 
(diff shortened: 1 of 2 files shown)


cls/SourceControl/Git/Extension.cls
```diff
@@ -191,7 +191,12 @@ Method OnSourceMenuContextItem(itemName As %String, menuItemName As %String, ByR
     } elseif ##class(SourceControl.Git.Utils).IsInSourceControl(itemName) {
         set Enabled = $case(menuItemName, "AddToSC":-1,:1)
     } else {
-        set Enabled = $case(menuItemName, "AddToSC":1,:-1)
+        if '$$$defClassDefined(classname) {
```
isc-tleavitt, line 194: Where is classname coming from? Need to see if InternalName is a .CLS and only do this check in that case.

isc-tleavitt (changes requested):
I don't think this will work - should hit <UNDEFINED> as is. Also needs CHANGELOG.md.

cls/SourceControl/Git/Extension.cls
isc-hwojnick: I initially had internal name in there and then deleted it because I abandoned that branch to work off of the one pravin made and then forgot to change classname to internalname 🤦‍♂️

isc-tleavitt: Will also need to strip the extension (.CLS)

isc-tleavitt (dismissed):
Thinking about this more, a few things I'd like to test:
What do we currently do if you try to add a whole package to source control? Will this break it? What about multiple selection of items? Does this work in VSCode or just Studio?
Also would be great if we could clean up the itemName parsing a little bit to make it more readable; Pravin pointed to `Type()` for this.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed JS errors in Studio on certain operations (#416)
+- Add menu option disabled for uncompiled files (#420)
```
isc-pbarton, line 12:
```suggestion
- Add menu option disabled for unsaved files (#420)
```

isc-pbarton (approved):

PR #421, opened 2024-07-10
tooltip now shows user who made uncommitted change

### LINKED ISSUE #411: Items owned by other users should show the username of the owning user in the tooltip
This is a minor regression from the UI revamp, but things are so much better now that it doesn't matter / shouldn't block release.

isc-hwojnick:
Fixes #411


isc-tleavitt (changes requested):
Code is good, needs CHANGELOG.md

Issue #433, opened 2024-07-16
Double quote characters (") not able to be used in commit messages


raymond-rebbeck:
Using a double quote character in a commit message results in an error: 
![double-quote-error](https://github.com/user-attachments/assets/57d6bdd6-f14c-4102-83e2-4d4a19f48093)

This is the commit message used to produce the above error, appears as though double quotes may not be getting escaped properly:
![test-commit](https://github.com/user-attachments/assets/f4ff5a8b-d132-446c-b5e8-a85b92c48161)

This is on latest latest main branch commit with d0b761ae3e2d25d48b0e3b788a36b1a11c7a55ce from PR #432 on top.

isc-pbarton:
PR #434 is a workaround that should be removed once we have a full fix for this issue.

isc-tleavitt:
From a design standpoint, I think we need a new API in WebUIDriver that allows arguments to be passed as an array and autoquoted by $zf(-100). We can start out using it for commit and eventually convert everything else over too.

isc-tleavitt:
And this should be part of the 2.4.1 release.

isc-hwojnick:
resolved through the new API

PR #434, opened 2024-07-16
Double quotes allowed in commit and amend messages

isc-hwojnick:
double quotes are turned into single quotes


raymond-rebbeck:
I don't think this actually fixes #433, you're still not really able to use double quotes in commit messages, they're just converted into something else, which I would consider more of a workaround than a fix.

isc-pbarton:
Agreed with Raymond that this is more of a workaround. Depending on how much time Henry has before the next release, we might merge the PR as is but keep the issue open for a full fix in a later release.

isc-pbarton (approved):

PR #435, opened 2024-07-17
Keep iris sync output

### LINKED ISSUE #426: On operations that sync IRIS with git repo, diagnostic output from sync is lost
Suppose syncing with IRIS (generally via SourceControl.Git.PullEventHandler:ForModifications or ForInternalNames) fails because of a compilation error - or succeeds with a bunch of stuff compiled. We have now way of knowing, because the output is thrown away, and any errors that occur are thrown as exceptions.

In SourceControl.Git.Utils:RunGitCommandWithInput we should allow the sync operation to fail. Output from the sync operation should be captured using the XDEV string buffer, same as is currently done in WebUIDriver (which will pose an issue - we'll need to remove that due to limitations with the string buffer and push everything into RunGitCommandWithInput). This output should be appended to the output stream with output from git, and if any errors occur they should be written to the err stream with error output from git. This should "just work" in the Web UI - errors will be shown as such.

Current bad behavior in the web ui is that after such a command errors, there's an output message that's totally useless saying that the webui server is not running - this is a holdout from the original Git WebUI open source project and doesn't make any sense at all because of course IRIS is running.

isc-hwojnick:
Fixes #426

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - WebUI workspace view now works properly for filenames with spaces (#423)
 - Fixed error popups in interop editors in Studio on 2024.1 (#417)
 - Reintroduced amend (#425)
-- IRIS sync output is no longer lost (#426)
+- Git operations that import items into IRIS now report output from compilation (#426)
 
 ## [2.4.0] - 2024-07-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - WebUI workspace view now works properly for filenames with spaces (#423)
 - Fixed error popups in interop editors in Studio on 2024.1 (#417)
 - Reintroduced amend (#425)
+- IRIS sync output is no longer lost (#426)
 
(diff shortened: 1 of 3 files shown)


cls/SourceControl/Git/WebUIDriver.cls
```diff
@@ -4,8 +4,6 @@ Class SourceControl.Git.WebUIDriver
 ClassMethod HandleRequest(pagePath As %String, InternalName As %String = "", Output handled As %Boolean = 0, Output %data As %Stream.Object)
 {
     // Make sure we capture any stray output
```
isc-pbarton, line 6: let's remove this comment which is no longer accurate

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - WebUI workspace view now works properly for filenames with spaces (#423)
 - Fixed error popups in interop editors in Studio on 2024.1 (#417)
 - Reintroduced amend (#425)
+- IRIS sync output is no longer lost (#426)
```
isc-pbarton, line 20:
```suggestion
- Git operations that import items into IRIS now report output from compilation (#426)
```

isc-pbarton (changes requested):
Just some nitpicks, removing a comment and suggesting a clearer description in changelog

isc-pbarton (approved):

PR #460, opened 2024-08-14
resolved issue with instance wide settings being in protected global

### LINKED ISSUE #444: Git commands fail for %Developer user if git not on path
If "git" is not on the path, and instead the full path to git is specified through git-source-control configuration, all git commands will fail for a user with just the %Developer role with a NOTOPEN.

The reason is that the "git bin path" configuration setting is stored in a global in the %SYS namespace, so you need access to that global. A %Developer will get a PROTECT error trying to access it.

isc-hwojnick:
Fixes #444


cls/SourceControl/Git/Utils.cls
```diff
@@ -183,6 +183,13 @@ ClassMethod IsImportAfter(menuItemName As %String) As %Boolean [ CodeMode = expr
 $Find(..#ImportAfterGitMenuItems, ","_menuItemName_",") > 0
 }
 
+ClassMethod MigrateInstanceSettings()
+{
+    if '$data(@..%SYSNamespaceStorage()@("%gitBinPath")) {
```
isc-pbarton, line 188: If there is no gitBinPath configured on the system, this will run every time git-source-control is installed, which could lead to weird behavior. I think instead we should check if any subscripts are defined in the old global (not just "%gitBinPath"). If yes, then we merge the old global to the new and kill the old global.

isc-pbarton (changes requested):

cls/SourceControl/Git/Utils.cls
```diff
@@ -183,6 +183,14 @@ ClassMethod IsImportAfter(menuItemName As %String) As %Boolean [ CodeMode = expr
 $Find(..#ImportAfterGitMenuItems, ","_menuItemName_",") > 0
 }
 
+ClassMethod MigrateInstanceSettings()
+{
+    if '$data(^["%SYS"]SYS("SourceControl", "Git")) {
```
isc-pbarton, line 188: Not quite there, the condition here is flipped. Make sure to test on an instance where ^["%SYS"]SYS("SourceControl","Git","%gitBinPath") is set.

isc-pbarton (changes requested):

isc-tleavitt (approved):

cls/SourceControl/Git/Utils.cls
isc-tleavitt: Oops - missed this in my quick read through. Good catch.

isc-pbarton (approved):

PR #479, opened 2024-09-16
Fixed Git web UI prompt to update file list

### LINKED ISSUE #478: Untick All on GIT GUI prompting will update files not selected that are not your changes will be pushed
We have a bunch of changes multiple users 
image
 If you select all then unselect all 
Then just have one file selected to push 
image

The Confirm comit still says it will send through other files

image

Some of these are a bit odd as cached error but TI is for instance a normal file 
image

isc-akadoo:
Fixes #478


isc-tleavitt (changes requested):
@isc-akadoo tricky thing on this - we have git-webui/src and git-webui/release, you need to make the change in /src and then zpm "load /path/to/git-source-control" will rebuild and propagate to the /release folder. So please make the change in /src as well, otherwise it'll get wiped out.

isc-tleavitt (approved):

PR #487, opened 2024-09-23
Fixed folder settings in mappings to be saved and persist

### LINKED ISSUE #483: Folder settings in mappings don't save consistently
Use case: set up mappings like the below:

image

Save them.

Change basic mode, then save again. Reload the page - the NoFolders status is lost from (at least) the display. Not sure if it's still present in the global.

isc-akadoo:
Fixes #483


isc-tleavitt (changes requested):
One change here.

Also, we need to bump the version to 2.5.0 in module.xml - could you take care of that as part of this PR? (Should have happened a long time ago - only just noticed the gap.)

PR #492, opened 2024-09-24
Add alerts when git is too old

### LINKED ISSUE #488: Behave better if user is running too old of a git version
On 2.17.0.windows.1 (which @isc-etamarch had running) the extension behaves horribly. Menu options don't show up and things randomly hang.

We should make the extension more helpful in this case.

isc-etamarch:
fixes #488

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.5.1] - Unreleased
+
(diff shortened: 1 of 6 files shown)


isc-tleavitt (changes requested):
I pulled the rug out from under this - sorry!

Please update version to 2.5.1 in module.xml and put the changelog in the "fixed" section for that new, unreleased version (ultimately I'd call this more of a bugfix as it corrects bad behavior of the extension in cases where it just wouldn't work)

Also, you need to run:
`zpm "git-source-control activate"`
To get the WebUI change rebuilt and moved to the right place (under the "release" folder which we also commit).

isc-tleavitt (approved):
@isc-etamarch it's good to put fixed issue numbers in CHANGELOG.md when relevant. I've added that for this case.

PR #494, opened 2024-09-24
Changed prompts in configure from 0/1 to no/yes

### LINKED ISSUE #461: Change prompts from 0/1 to no/yes
The new prompts around basic mode are a little confusing - instead of saying 0/1 they should say yes/no (we should use the proper %Library.Prompt for this)

isc-akadoo:
Fixes #461


isc-tleavitt (changes requested):
Basic mode is tricky - to match the UI behavior it needs to be treated a bit differently, with 0/1/empty string as options (where empty string means "follow the system default behavior"). Using GetMenu / GetArray from %Library.Prompt would be better here.

isc-tleavitt (approved):

PR #502, opened 2024-09-30
Added import-all to basic mode

### LINKED ISSUE #498: Make Import All options available in Basic mode
Consider enabling the Import All and Import All (Force) options in Basic mode. Currently these are disabled because they are not a standard part of the basic mode workflow, but they are useful to have when something goes wrong and IRIS is out of sync with the repository.

isc-etamarch:
Fixes #498

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added 
 - Files in uncommitted queue in any namespace warn users when opened except for in VSCode (#370)
-
-### Added
 - Added link back to IRIS management portal from Settings, Git WebUI pages (#449)
+- Added Import all and Import All (Force) to basic mode menu
 
 ### Fixed
 - Changed prompts in configure from 0/1 to no/yes (#461)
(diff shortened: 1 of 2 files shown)


isc-tleavitt:
@isc-pbarton do you think there's any benefit in having this with/without the Force flag?

isc-pbarton:
On second thought I like having just the Import All option and leaving out Import All (Force) because Force is only useful in more advanced debugging scenarios. But I don't have an extremely strong opinion about it.

isc-tleavitt:
I thought about this a bunch and discussed with @isc-pbarton and ultimately this is the right thing. We talked about renaming the menu items but the current names make more intuitive sense than anything else we could come up with. There are definitely reasonable use cases for both, even in basic mode. So we'll keep this as-is.

CHANGELOG.md
```diff
@@ -9,9 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added 
 - Files in uncommitted queue in any namespace warn users when opened except for in VSCode (#370)
-
-### Added
 - Added link back to IRIS management portal from Settings, Git WebUI pages (#449)
+- Added Import all and Import All (Force) to basic mode menu
```
isc-tleavitt, line 13: Nitpick: please add reference to original issue number in changelog

isc-tleavitt (changes requested):

isc-tleavitt (approved):

PR #504, opened 2024-09-30
Use merge instead of rebase to sync other developers' changes in basic mode

### LINKED ISSUE #503: Merging default remote branch does not import changed code
In Basic mode, if I Sync and have a default merge branch configured, changes rebased from the default merge branch do not actually get imported and compiled into IRIS.
The way we are diffing across the rebase to list files that have changed never shows any changes.
The reason we never caught this before is because we would test by opening the changed item in VS Code, which calls the OnBeforeLoad to load the item from disk.

isc-pbarton:
Fixes #503 
The logic we were using previously to list files that needed to be imported after a rebase was broken. This change fixes it by using a merge instead of a rebase. We originally went with rebase because the git commit history looks nicer that way, but I think merge is going to be a lot simpler to support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added 
 - Files in uncommitted queue in any namespace warn users when opened except for in VSCode (#370)
-
-### Added
 - Added link back to IRIS management portal from Settings, Git WebUI pages (#449)
 
 ### Fixed
 - Changed prompts in configure from 0/1 to no/yes (#461)
 - Added warnings when user is using incompatible git version (#488)
 - Fixed the back button navigation between WebUI and Settings page (#361)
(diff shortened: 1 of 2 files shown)


isc-tleavitt (changes requested):
@isc-pbarton we're going to need a corresponding update to SourceControl.Git.Util.ProductionConflictResolver:Resolve to change rebase --continue to merge --continue.

In general I prefer the idea of rebasing over merging from a history/tracking perspective, especially for cases where we're using the deployment manager on top of the main branch; merges are going to end up being deceptive. Is it possible to just fix the diff command?

isc-pbarton:
That's fair, I'll look into fixing the diff instead.

raymond-rebbeck:
I have not used basic mode at all but I too prefer rebasing over merging or at the very least having a choice.

isc-pbarton:
Going to close this out and create a new merge request with the rebase fix.

PR #506, opened 2024-10-01
Rebasing from remote loads code changed on the remote

### LINKED ISSUE #503: Merging default remote branch does not import changed code

isc-pbarton:
Fixes #503 
We now diff the old head with the new head. Before we diffed the remote branch with the new head, which misses all the changes made on the remote branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added 
 - Files in uncommitted queue in any namespace warn users when opened except for in VSCode (#370)
-
-### Added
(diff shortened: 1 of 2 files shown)


isc-tleavitt (changes requested):
Looks good - just needs changelog. Glad this was simple!!

isc-tleavitt (approved):

Issue #495, opened 2024-09-25
Possible issues with differing file extensions on HL7 between VSCode and Management Portal


isc-tleavitt:
I've observed in my own testing and in review of a customer call recording some possible differences in behavior / issues related to HL7 file extensions being inconsistent with editing from different places. Because of OS-level differences this is something that would only end up mattering on Linux, of course.

I think the root cause / solution might end up being similar to https://github.com/intersystems/ipm/issues/527

isc-tleavitt:
The issue seems to be that .HL7 (upper case) isn't properly recognized as the right document type.

isc-etamarch:
I've noticed that the internal name mostly used .CLS (sometimes .cls), while external name is usually .cls. Probably not causing the HL& issues as you noted above, but could also potentially cause issues

PR #508, opened 2024-10-01
Fix hl7 extensions inconsistencies

isc-etamarch:
Addresses #495


isc-tleavitt (approved):
Reviewed this with @isc-etamarch in a few calls today. There are a few other things we'll want to do but this is good for purposes of the current issue/PR.

@isc-pbarton I'd appreciate if you could review too since I made a few of the changes here.

PR #511, opened 2024-10-02
Improve handling of truncated class names from git pull

raymond-rebbeck:
The current way of handling truncated class names from git pull seems to have two major flaws:
1. It is possible for truncated class names to fail to be expanded if the repository directory structure does not align with class packages
2. When this failure occurs and no possible classes are found, there is no warning generated

For example if the repository has the following base directory structure:
```/foo/bar/``` and contains class ```A.B.C``` stored on disk as ```A/B/C.cls``` (where foo and bar have no relation to any class package names)

then git pull may provide truncated output such as the following:
```.../bar/A/B/C.cls```

```ExpandClasses``` will try and fail to find a class named ```bar.A.B.C``` with no warning generated, the class silently fails to import.

```ExpandClasses``` has been updated so that upon failing to find any class names it will remove period-delimited pieces from the start of the name and try again, until it finds something or runs out of pieces.

This would mean for example that after failing to find ```bar.A.B.C``` it would then try ```A.B.C``` and find the intended class

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the back button navigation between WebUI and Settings page (#361)
 - Basic mode Sync operation now imports items changed on the remote merge branch (#506)
 - Fetch diff output uses correct remote branch (#509)
+- Properly handle more cases of truncated filenames from git pull (#511)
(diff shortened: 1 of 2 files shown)


isc-tleavitt:
@raymond-rebbeck another great catch - thank you!

isc-tleavitt (approved):

PR #512, opened 2024-10-02
Improved behavior for commits when attribution settings are not configured

### LINKED ISSUE #450: Behave better if attribution settings are not configured
Currently, if no attribution settings are configured, commit will fail silently (through the menu) or loudly (through the UI).

We should drive the user to correct action in all of these cases.

isc-akadoo:
Fixes #450


cls/SourceControl/Git/Utils.cls
```diff
@@ -262,6 +262,7 @@ ClassMethod UserAction(InternalName As %String, MenuName As %String, ByRef Targe
         set Action = 7
         quit $$$OK
     } elseif (menuItemName = "Commit") {
+        do ..CheckUserIdentity()
```
isc-tleavitt, line 265: We can do better than outputting warnings - if user identity is not configured, the commit *will* fail.

See https://docs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25Studio.Extension.Base#UserAction - we should display an alert dialog rather than a prompt if the identity isn't configured. We should do the same for the Sync menu item as it will also subsequently fail.

cls/SourceControl/Git/Utils.cls
```diff
@@ -1784,6 +1793,15 @@ ClassMethod RunGitCommandWithInput(command As %String, inFile As %String = "", O
     set errStream = ##class(%Stream.FileCharacter).%OpenId(errLog,,.sc)
     set outStream = ##class(%Stream.FileCharacter).%OpenId(outLog,,.sc)
     set outStream.TranslateTable="UTF8"
+
+    if ((isCommit) && (returnCode = 128)){
+        set errStreamLine = errStream.ReadLine()
+        if (((errStreamLine = "Committer identity unknown") || (errStreamLine = "Author identity unknown"))) {
+            do errStream.Clear()
+            do errStream.WriteLine("Commit failed as user identity unknown."_$c(13,10)_$c(13,10)_"Go to settings to configure Git committer name and email."_$c(13,10))
```
isc-tleavitt, line 1801: Let's be a bit more verbose, actually (here and elsewhere):
"Go to the Settings page"

isc-tleavitt (changes requested):
See feedback; also, as I'd said via Teams: It would also be helpful to produce some output on initial connection (via VSCode/Studio - this is the Type = 1, Name = 5 block in SourceControl.Git.Extension:UserAction) to provide an earlier warning.

cls/SourceControl/Git/Extension.cls
```diff
@@ -49,7 +49,11 @@ Method UserAction(Type As %Integer, Name As %String, InternalName As %String, Se
     if Type = 1, Name = 5 {
         // reroute to Status menu option
         set Name = "%SourceMenu,Status"
-    } 
+        if (settings.gitUserName = "") || (settings.gitUserEmail = "") {
+            set Target = "Warning: Git committer name or email is not configured. Go to settings."
```
isc-tleavitt, line 53: Clarifying - the wording re: "settings" should be:
"Go to the Settings page to fix this."
Or maybe:
"Go to the Settings page in the Source Control menu to fix this."

The important thing is that we're pointing to "the Settings page" rather than "settings" which seems ambiguous.

cls/SourceControl/Git/Extension.cls
```diff
@@ -49,7 +49,11 @@ Method UserAction(Type As %Integer, Name As %String, InternalName As %String, Se
...
+            set Action = 6
```
isc-tleavitt, line 54: We want to take this same approach not just on initial connection but also on the Commit and Sync menu items. If attribution settings aren't set up, in any of these cases, we want to change Action to 6 (show an alert), put the same helpful message in Target, and not try to Commit or Sync (rather than trying to and failing later in the process). This is especially relevant for Sync.

If you want to refactor this to a helper method, that's great. It's also good to have the detection in RunGitCommandWithInput specifically, because we also want to be helpful for commits via the Git WebUI.

isc-tleavitt (changes requested):

isc-tleavitt (approved):

PR #514, opened 2024-10-03
Fix for JS syntax error preventing use of Web UI with Studio

raymond-rebbeck:
Web UI was broken in Studio by b0ff1430


isc-tleavitt (approved):

isc-tleavitt:
Skipping changelog as this fixes an unreleased bug.

@isc-etamarch @isc-pbarton new plan: for larger UI changes we should test in Studio on the PR rather than possibly waiting for pre-release testing.

PR #515, opened 2024-10-03
Make expanding truncated filenames more general and add support for lookup tables

raymond-rebbeck:
Makes expanding truncated file names from git pull a bit more general and adds support for handling lookup tables. The same concept is applied to lookup tables as was previously introduced for classes in #511 

After resetting my local branch back to the first ever commit I can use this to do a pull and successfully  import a codebase containing 600+ lookups with names as long as 56 characters.

### DIFF
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -1885,22 +1885,18 @@ ClassMethod SyncIrisWithRepoThroughCommand(ByRef outStream) As %Status
             set externalName = $zstrip($piece(line,"|",1),"<>W")
             if $Extract(externalName,1,3) = "..." {
                 // For extremely long file names, git may truncate the path.
-                // Simplifying assumption: this is a class, because nothing else would have that long a name.
-                // In other cases, we'll just end up logging the invalid externalName.
-                if $Piece(externalName,".",*) = "cls" {
-                    set possibleClasses = ..ExpandClasses(externalName)
-                    if $ListLength(possibleClasses) '= 0 {
-                        set pointer = 0
-                        while $ListNext(possibleClasses,pointer,class) {
-                            set modification = ##class(SourceControl.Git.Modification).%New()
-                            set modification.changeType = "C"
-                            set modification.internalName = class_".CLS"
-                            set modification.externalName = ..ExternalName(modification.internalName)
-                            set files($i(files)) = modification
-                        }
-                    } else {
-                        write !,"WARNING: unable to translate external name ",externalName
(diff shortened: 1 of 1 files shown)


cls/SourceControl/Git/Utils.cls
```diff
@@ -1930,28 +1926,42 @@ ClassMethod SyncIrisWithRepoThroughCommand(ByRef outStream) As %Status
...
+        set internalName = $Translate(internalName,"%",..PercentClassReplace())
+    }
+    
+    set possibleNames = ""
     do {
-        &sql(select %DLIST(Name) into :classes from %Dictionary.ClassDefinition where Name like '%'||:internalName)
+        if (externalNameExtension = "cls") {
+            &sql(select %DLIST(Name) into :possibleNames from %Dictionary.ClassDefinition where Name like '%'||:internalName)
+        } elseif (externalNameExtension = "lut") {
+            &sql(select %DLIST(distinct TableName) into :possibleNames from Ens_Util.LookupTable where TableName like '%'||:internalName)
```
isc-tleavitt, line 1945: I think some additional work might be needed to generalize this. Two major reasons:
* We want this to work on non-interoperability-enabled namespaces, where Ens_Util.LookupTable isn't available. This is a blocking issue for merge of this PR.
* We want this to work for other document types as well, not just cls and lut. (e.g., I could also see truncation popping up with with .DFI)

There might be an elegant (code-wise) approach using %Library.RoutineMgr:StudioOpenDialog that would hit all relevant cases at once, but I'd be concerned about the possible performance impact. Another approach could be using the [List query of the relevant %Studio.AbstractDocument subclass](https://docs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25Studio.AbstractDocument#List) directly based on the extension.

isc-tleavitt (changes requested):
@raymond-rebbeck thank you for yet another helpful PR! I need to think on this one a bit before merging; it will need at least minor changes but might make sense to further generalize for non-class (and LUT) document types while we're at it.

isc-tleavitt:
Closing this in favor of #517

cls/SourceControl/Git/Utils.cls
raymond-rebbeck: Thank you for the excellent feedback. Based on this I am thinking that attempting to address this by improving the truncated name expanding logic may not be the ideal approach and I have instead put together #517 to hopefully address this more robustly, at least for pulls initially.

PR #518, opened 2024-10-04
Convert URLs in Sync output UI into clickable links

### LINKED ISSUE #497: Convert URLs in Sync interface output into clickable links
This is especially helpful when sync pushes changes to the remote, and the remote responds with some output saying "go to this URL to create a merge request".
We already do this in output from the Git Web UI, see #377

isc-akadoo:
Fixes #497

### DIFF
--- a/csp/sync.csp
+++ b/csp/sync.csp
@@ -147,15 +147,8 @@
             &js<
                 var outputContainer = document.getElementById('outputBox');
                 var lineText = #(..QuoteJS(escapedLine))#;
-                var lineTextAsArr = lineText.split(" ");
-                lineTextAsArr = lineTextAsArr.map(function(text){
-                    if (text.substring(0,8) === "https://") {
-                        return '<a href="' + text + '" target="_blank">' + text + '</a>';
-                    } else {
-                        return text;
-                    }
-                });
-                lineText = lineTextAsArr.join(" ")
+                var urlRegex = /(https?:\/\/[^ ]*)/;
+                lineText = lineText.replace(urlRegex, "<a href='$1' target=_blank>$1</a>");
                 var lineTextNode = document.createTextNode(lineText);
                 outputContainer.innerHTML += lineText + "<br>";
             >
--- a/csp/sync.csp
+++ b/csp/sync.csp
@@ -147,9 +147,8 @@
             &js<
                 var outputContainer = document.getElementById('outputBox');
                 var lineText = #(..QuoteJS(escapedLine))#;
-                var urlRegex = /(https?:\/\/[^ ]*)/;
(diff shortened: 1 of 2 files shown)


csp/sync.csp
```diff
@@ -147,6 +147,15 @@
             &js<
                 var outputContainer = document.getElementById('outputBox');
                 var lineText = #(..QuoteJS(escapedLine))#;
+                var lineTextAsArr = lineText.split(" ");
+                lineTextAsArr = lineTextAsArr.map(function(text){
+                    if (text.substring(0,8) === "https://") {
+                        return '<a href="' + text + '" target="_blank">' + text + '</a>';
```
isc-tleavitt, line 153: I think there's minor potential for DOM-based XSS with this approach, but practically speaking since we're handling output from the git remote if someone wanted to exploit this they'd be able to do much easier and worse things.

Ref: https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#rule-2-javascript-escape-before-inserting-untrusted-data-into-html-attribute-subcontext-within-the-execution-context

You could probably do the same in a lot less code (would also get around the map issue) with a regex replace - see e.g. https://stackoverflow.com/questions/31760030/extracting-for-url-from-string-using-regex .

isc-tleavitt (changes requested):

csp/sync.csp
isc-tleavitt: Tweaking the regex a bit would also provide a simple mitigation for the XSS risk

isc-tleavitt: Well... actually, the XSS risk is already there. So we'll worry about that separately.

isc-tleavitt: e.g., use this regex:
/(https?:\/\/[^ "]*)/gi

csp/sync.csp
```diff
@@ -147,6 +147,8 @@
             &js<
                 var outputContainer = document.getElementById('outputBox');
                 var lineText = #(..QuoteJS(escapedLine))#;
+                var urlRegex = /(https?:\/\/[^ ]*)/;
```
isc-tleavitt, line 150: Change to /(https?:\/\/[^ ]*)/gi

csp/sync.csp
```diff
@@ -147,6 +147,8 @@
...
+                lineText = lineText.replace(urlRegex, "<a href='$1' target=_blank>$1</a>");
                 var lineTextNode = document.createTextNode(lineText);
```
isc-tleavitt, line 152: Kill this line while you're at it - we don't actually use lineTextNode.

isc-tleavitt (changes requested):

isc-tleavitt (approved):

PR #520, opened 2024-10-04
Added Environment

### LINKED ISSUE #124: Add equivalent to $$Env^%buildccr
git-source-control should have APIs similar to $$Env^%buildccr, $$IsLIVE^%buildccr (to support different stages for environments)

isc-etamarch:
Ask for environment in configuration; Add changing environment in settings; Show environment in UI

fixes (#124 )

### DIFF
--- a/cls/SourceControl/Git/Settings.cls
+++ b/cls/SourceControl/Git/Settings.cls
@@ -182,8 +182,9 @@ ClassMethod Configure() As %Boolean [ CodeMode = objectgenerator ]
             do %code.WriteLine(" set list(1) = ""DEVELOPMENT""")
             do %code.WriteLine(" set list(2) = ""TEST""")
             do %code.WriteLine(" set list(3) = ""LIVE""")
-            do %code.WriteLine(" set list(4) = """"")
-            do %code.WriteLine(" set value = """"")
+            do %code.WriteLine(" set list(4) = ""FAILOVER""")
+            do %code.WriteLine(" set list(5) = """"")
+            do %code.WriteLine(" set value = inst.property")
             do %code.WriteLine(" set response = ##class(%Library.Prompt).GetArray("_promptQuoted_",.value,.list,,,,"_defaultPromptFlag_")")
         } else {
             do %code.WriteLine(" set response = ##class(%Library.Prompt).GetString("_promptQuoted_",.value,,,,"_defaultPromptFlag_")")
--- a/cls/SourceControl/Git/Settings.cls
+++ b/cls/SourceControl/Git/Settings.cls
@@ -119,8 +119,12 @@ Method %Save() As %Status
     } else {
         set @storage@("settings", "user", $username, "basicMode") = ..basicMode
     } 
+    try {
+        do $SYSTEM.Version.SystemMode(..environmentName)
+    } catch e {
+        // no-op; user might not have privileges.
+    }
 
-    do $SYSTEM.Version.SystemMode(..environmentName)
     
 
     // update value of basicUserMode to reflect the updated setting for basicMode
@@ -184,7 +188,6 @@ ClassMethod Configure() As %Boolean [ CodeMode = objectgenerator ]
             do %code.WriteLine(" set list(3) = ""LIVE""")
             do %code.WriteLine(" set list(4) = ""FAILOVER""")
             do %code.WriteLine(" set list(5) = """"")
-            do %code.WriteLine(" set value = inst.environmentName")
             do %code.WriteLine(" set response = ##class(%Library.Prompt).GetArray("_promptQuoted_",.value,.list,,,,"_defaultPromptFlag_")")
         } else {
             do %code.WriteLine(" set response = ##class(%Library.Prompt).GetString("_promptQuoted_",.value,,,,"_defaultPromptFlag_")")
--- a/cls/SourceControl/Git/Settings.cls
+++ b/cls/SourceControl/Git/Settings.cls
@@ -54,7 +54,7 @@ Property namespaceLevelGitWebApp As %Boolean [ InitialExpression = {##class(Sour
 Property warnInstanceWideUncommitted As %Boolean [ InitialExpression = {##class(SourceControl.Git.Utils).WarnInstanceWideUncommitted()} ];
 
 /// The name of the environment (DEVELOPMENT, TEST, LIVE)
-Property environmentName As %String(MAXLEN = "", VALUELIST = ",DEVELOPMENT,TEST,LIVE") [ InitialExpression = {##class(SourceControl.Git.Utils).EnvironmentName()} ];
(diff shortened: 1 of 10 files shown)


cls/SourceControl/Git/Settings.cls
```diff
@@ -110,12 +113,16 @@ Method %Save() As %Status
...
     set @storage@("settings", "warnInstanceWideUncommitted") = ..warnInstanceWideUncommitted
     set @storage@("settings", "basicMode") = ..systemBasicMode
+    set @storage@("settings", "environmentName") = ..environmentName
     if ..basicMode = "system" {
         kill @storage@("settings", "user", $username, "basicMode")
     } else {
         set @storage@("settings", "user", $username, "basicMode") = ..basicMode
     } 
 
+    set ^%SYS("SystemMode") = ..environmentName
```
isc-pbarton, line 123: Nitpick: might prefer calling $SYSTEM.Version.SystemMode(environment name) instead of setting the global directly.

cls/SourceControl/Git/Utils.cls
```diff
@@ -163,6 +163,16 @@ ClassMethod WarnInstanceWideUncommitted() As %Boolean
     quit $get(@..#Storage@("settings","warnInstanceWideUncommitted"),1)
 }
 
+ClassMethod EnvironmentName() As %String
+{
+    quit $get(@..#Storage@("settings","environmentName"),1)
```
isc-pbarton, line 168: Let's default to the empty string instead of 1 for the environment name

isc-pbarton (dismissed):

cls/SourceControl/Git/Utils.cls
```diff
@@ -163,6 +163,16 @@ ClassMethod WarnInstanceWideUncommitted() As %Boolean
     quit $get(@..#Storage@("settings","warnInstanceWideUncommitted"),1)
 }
 
+ClassMethod EnvironmentName() As %String
+{
+    quit $get(@..#Storage@("settings","environmentName"),"")
```
isc-tleavitt, line 168: IMO this shouldn't be in a separate global - let's just have this call $System.Version.SystemMode

cls/SourceControl/Git/Utils.cls
```diff
@@ -163,6 +163,16 @@ ClassMethod WarnInstanceWideUncommitted() As %Boolean
...
+}
+
+ClassMethod IsLIVE() As %Boolean
+{
+    quit $get(@..#Storage@("settings","environmentName"),1)="LIVE"
```
isc-tleavitt, line 173: Should call ..EnvironmentName() to reduce code duplication

cls/SourceControl/Git/Settings.cls
```diff
@@ -171,6 +178,13 @@ ClassMethod Configure() As %Boolean [ CodeMode = objectgenerator ]
             do %code.WriteLine(" if value = ""System's Default Mode"" { set value = ""system""}")
             do %code.WriteLine(" elseif value = ""Yes"" { set value = 1}")
             do %code.WriteLine(" elseif value = ""No"" { set value = 0}")
+        } elseif ((propertyDef) && (propertyDef.Name = "environmentName")) {
+            do %code.WriteLine(" set list(1) = ""DEVELOPMENT""")
+            do %code.WriteLine(" set list(2) = ""TEST""")
+            do %code.WriteLine(" set list(3) = ""LIVE""")
```
isc-tleavitt, line 184: Per https://docs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25SYSTEM.Version#SystemMode "FAILOVER" is also an option - should add that.

cls/SourceControl/Git/Settings.cls
```diff
@@ -171,6 +178,13 @@ ClassMethod Configure() As %Boolean [ CodeMode = objectgenerator ]
...
+            do %code.WriteLine(" set list(4) = """"")
+            do %code.WriteLine(" set value = """"")
```
isc-tleavitt, line 186: Default value should be the current system mode

cls/SourceControl/Git/Settings.cls
```diff
@@ -110,12 +113,16 @@ Method %Save() As %Status
...
     set @storage@("settings", "warnInstanceWideUncommitted") = ..warnInstanceWideUncommitted
     set @storage@("settings", "basicMode") = ..systemBasicMode
+    set @storage@("settings", "environmentName") = ..environmentName
     if ..basicMode = "system" {
         kill @storage@("settings", "user", $username, "basicMode")
     } else {
         set @storage@("settings", "user", $username, "basicMode") = ..basicMode
     } 
 
+    do $SYSTEM.Version.SystemMode(..environmentName)
```
isc-tleavitt, line 123: Please specifically test that someone with just %Developer can set this

isc-tleavitt (changes requested):
My review is up - a few changes recommended

cls/SourceControl/Git/Settings.cls
```diff
@@ -182,8 +182,9 @@ ClassMethod Configure() As %Boolean [ CodeMode = objectgenerator ]
             do %code.WriteLine(" set list(1) = ""DEVELOPMENT""")
             do %code.WriteLine(" set list(2) = ""TEST""")
             do %code.WriteLine(" set list(3) = ""LIVE""")
-            do %code.WriteLine(" set list(4) = """"")
-            do %code.WriteLine(" set value = """"")
+            do %code.WriteLine(" set list(4) = ""FAILOVER""")
+            do %code.WriteLine(" set list(5) = """"")
+            do %code.WriteLine(" set value = inst.property")
```
isc-tleavitt: This is wrong - should be `do %code.WriteLine(" set value = inst."_property)` - although actually, shouldn't be needed at all because we do that on line 162.

isc-tleavitt (changes requested):

cls/SourceControl/Git/Settings.cls
isc-tleavitt: @isc-etamarch we overlapped on this - you were fixing as I commented, but my point still stands that we shouldn't need to set value here at all.

cls/SourceControl/Git/Settings.cls
isc-etamarch: $SYSTEM.Version.SystemMode does not work for someone with %Developer

isc-etamarch: Should I revert to setting the global directly (through a private helper method)?

isc-tleavitt: Setting the global directly isn't going to work either - I'd recommend just wrapping the whole thing in a Try/Catch and ignoring any errors.

isc-tleavitt: @isc-etamarch Yes, it'll just be a no-op.

isc-tleavitt (approved):

PR #527, opened 2024-10-08
Add an option to allow git push --force to be performed in expert mode

raymond-rebbeck:
Add **Push to remote branch (Force)** to the menu, allowing an expert mode user to perform ```git push --force```

The user is prompted with a warning and to confirm that they want to do this.

The motivation for this is that a basic mode user is capable of indirectly performing a force push via the **Sync** option, potentially without even realising or understanding that they are doing so. An expert mode user should be able to do the same with intent but currently has no option available outside of manually interacting with the local repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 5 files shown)


isc-tleavitt:
@isc-etamarch could you tack on appropriate updates to module.xml and CHANGELOG.md for this (given that we released yesterday)? It's a good simple change and quite reasonable, happy to merge as-is with those updates.

In CHANGELOG.md, for cases like this where there's a PR but no issue (because the PR describes the issue), just put the PR number in the change note.

isc-etamarch (approved):

PR #536, opened 2024-10-10
no Commits on default merge branch in basic mode

### LINKED ISSUE #484: In basic mode, prohibit commits to default merge branch
If we are in basic mode and have a default merge branch configured (call it "development"), there should never be commits from git-source-control directly into development. A couple of scenarios where this could happen and lead to issues:

- I switched back to the development branch after finishing a feature and started working on a new feature but forgot to create a new feature branch
- I tried to create a new feature branch but it failed for some reason. I assume I'm on the new branch because I didn't check source control output, but I'm actually on development.

If the current branch is development, the sync option should not add, commit, or push. It should just pull.

isc-etamarch:
Fixes #484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed display of other users' username in workspace view on Unix (#530)
+- Prevent direct commits to default merge branch in basic mode (#484)
 
 ## [2.6.0] - 2024-10-07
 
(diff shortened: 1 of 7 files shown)


isc-tleavitt (changes requested):
You can also commit a single modified item from the menu in basic mode - this should also be prevented.

isc-tleavitt (changes requested):
One more thought on this: we need to consider usability for a user in basic mode trying to commit/sync. Hiding the "Commit" option and producing different output on the "Sync" option doesn't help our less-technical users. I'd rather see "Commit" enabled (so the user realizes from the menu they have uncommitted changes to the current item) and provide an alert saying they need to create a new branch to commit changes against. Similarly, on Sync, we should be really clear to the user that their work *won't* be committed - e.g., a big red warning box.

isc-etamarch:
Added some alerts:

image
image

PR #537, opened 2024-10-10
Compact sidebar

### LINKED ISSUE #525: Revisit left-side navigation spacing in WebUI for usability on smaller monitors
Zooming helps, but with a large number of branches the left-side navigation is becoming less usable as we add more options. Consider compacting spacing.

isc-etamarch:
Fixes #525


isc-etamarch:
@isc-tleavitt Let me know if we need further compacting / other UI changes to the sidebar

isc-tleavitt (changes requested):
@isc-etamarch this is a vast improvement, thank you.

My main remaining concern is the behavior with highly limited vertical space - there might not be enough for the remote branch view. This is great:
image

But this is not - from in VSCode (which is particularly egregiously bad for some reason - maybe there are other steps we could take re: zoom/etc. to make this better):
image

Note the overlap of "Remote Branches" with the links underneath, and the fact that you cant see "More..."

I don't want to go too crazy on this, but let's try to make it a bit better first.

isc-tleavitt (approved):

Issue #510, opened 2024-10-02
Provide a way to add favorites for just the current user/namespace via Settings page


isc-tleavitt:
It would be great to have a button in Settings to add the favorites links for the current user and namespace. This is especially relevant for environments like Health Connect Cloud where new users and namespaces will be added after git-source-control is installed. Note: implementation will probably require role escalation via a Z* / IRISSYS class (preferred) or privileged routine application.

This is related to #468 and could make sense to implement at the same time, as we could also use this button to clean up / correct favorites links for the current user.

PR #538, opened 2024-10-10
 Simplify favorites and landing page for multi-namespace setup

### LINKED ISSUE #468: Simplify favorites and landing page for multi-namespace setup
The common pattern in Health Connect Cloud is to have multiple dev namespaces. Our current git-source-control installation process creates a favorite link to each namespace, but really we should just create a single link and either (a) use the request context namespace, if possible, to determine namespace selection or (b) allow the user to drop in to the Git WebUI with no namespace context and support picking one right away, with a default selection of a namespace mapping to the current user's username if such a namespace exists. Really, even if (a) works, we should have support for (b) as well.

isc-akadoo:
Fixes #468, #510


cls/SourceControl/Git/Utils.cls
```diff
@@ -72,6 +72,11 @@ ClassMethod SettingsUIReadOnly() As %Status [ CodeMode = expression ]
 $Get(@..#Storage@("settings","settingsUIReadOnly"), 0)
 }
 
+ClassMethod FavoriteNamespaces() As %String
+{
+    quit $get(@..#Storage@("settings", "user", $username, "favNamespaces"),"")
```
isc-tleavitt, line 77: Ideally this would be calculated using SQL rather than stored (which as implemented means storing separately in all namespaces even though it's inherently a system-wide setting - not really ideal).

cls/SourceControl/Git/WebUIDriver.cls
```diff
@@ -148,7 +148,7 @@ ClassMethod HandleRequest(pagePath As %String, InternalName As %String = "", Out
                 do discardsInBranch.%ToJSON(%data)
                 set handled = 1
             } elseif (pathStart = "contexts") {
-                set contexts = ##class(SourceControl.Git.Utils).GetContexts()
+                set contexts = ##class(SourceControl.Git.Utils).GetContexts(1)
```
isc-tleavitt, line 151: I think this change also impacts the WebUI - could make this subject to a URL parameter so the WebUI continues to function as-is but we don't try to git pull in non-namespace contexts.

csp/gitprojectsettings.csp
```diff
@@ -550,6 +550,35 @@ body {
             </div> -->
         </div>
 
+        <div class="form-group row mb-3">
+            <label for="addToFav" class="offset-sm-1 col-sm-3 col-form-label"  data-toggle="tooltip" data-placement="top" title="Favorite namespaces for user. For Windows/Mac: Hold down the control(ctrl)/command(cmd) button to select multiple options.">Favorite Namespaces</label>
+            <div class="col-sm-7">
+                <select multiple class="form-control" id="addToFav" name="favNamespace">
+                <server>
+                    set contextList = ##class(SourceControl.Git.Utils).GetContexts(1)
```
isc-tleavitt, line 558: Idea: you could have SourceControl.Git.Utils return the set of contexts as a byref local array, that would let you take a more sensible approach below.

csp/pull.csp
```diff
@@ -45,6 +56,13 @@ function execute() {
     }
 }
 
+function updateContext() {
+    const contextSelect = document.getElementById('newContext');
```
isc-tleavitt, line 60: I'd recommend just using "var" instead of "const" (in case we ever end up using this from Studio)

CHANGELOG.md
```diff
@@ -10,9 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added 'git push --force' in expert mode (#527)
 - Add remote repository to settings page (#448)
+- Added change context option to pull page (#468)
+- Added favorite namespaces setting for a user (#468, #510) 
 
 ### Fixed
 - Fixed display of other users' username in workspace view on Unix (#530)
+- Fixed GetContexts utils function to return namespaces for WebUI(#468)
```
isc-tleavitt, line 18: I'd rephrase to say "Don't include implied namespaces in list of namespaces"

isc-tleavitt (changes requested):
A few points of feedback; also, in general, please be sure to test as %Developer. If you need additional pointers to using the HCC environment for this please let me know.

cls/SourceControl/Git/Utils.cls
```diff
@@ -2832,4 +2843,50 @@ ClassMethod SetConfiguredRemote(url) As %String
     quit output
 }
 
+ClassMethod ConfigureFavoriteNamespaces(username As %String, newNamespaces As %String)
+{
+    // Delete all the GIT favorite links for the user
+    set iterator = newNamespaces.%GetIterator()
```
isc-etamarch, line 2849: Need to use embedded sql here to avoid privilege checking, since %Developer don't have access to the %SYS_Portal.Users table

cls/SourceControl/Git/Utils.cls
```diff
@@ -2832,4 +2843,50 @@ ClassMethod SetConfiguredRemote(url) As %String
...
+    while iterator.%GetNext(.key, .value) {
+        set sql = "delete from %SYS_Portal.Users where Username='"_username_"' and Page like '%Git%'"
+        set statement = ##class(%SQL.Statement).%New()
+        do ##class(%SQL.Statement).%ExecDirect(statement,sql)
+    }
+
+    set iterator = newNamespaces.%GetIterator()
+    while iterator.%GetNext(.key, .value) {
```
isc-etamarch, line 2857: This also needs to be embedded SQL, does not work for %Developer

cls/SourceControl/Git/Utils.cls
```diff
@@ -2832,4 +2843,50 @@ ClassMethod SetConfiguredRemote(url) As %String
...
+        do ##class(%SQL.Statement).%ExecDirect(statement,sql)
+    }
+}
+
+ClassMethod GetFavoriteNamespaces(ByRef favNamespaces As %DynamicArray, ByRef nonFavNamespaces As %DynamicArray) {
+    set allNamespaces = ..GetContexts(1)
+    set iterator = allNamespaces.%GetIterator()
+
+    while iterator.%GetNext(.key, .value) {
+        set sql = "select * from %SYS_Portal.Users where username='"_$username_"' and page='Git: "_value_"'"
```
isc-etamarch, line 2881: embedded sql here too

csp/gitprojectsettings.csp
```diff
@@ -564,6 +576,26 @@ body {
             </div> -->
         </div>
 
+        <div class="form-group row mb-3">
+            <label for="addToFav" class="offset-sm-1 col-sm-3 col-form-label"  data-toggle="tooltip" data-placement="top" title="Favorite namespaces for user. For Windows/Mac: Hold down the control(ctrl)/command(cmd) button to select multiple options.">Favorite Namespaces</label>
```
isc-etamarch, line 580: The instruction to use should be visible without hover (maybe below / above the box), since usage is not fully intuitive

isc-etamarch (changes requested):
We need to make sure this works for %Developer by using embedded SQL for all SQL statements

cls/SourceControl/Git/Utils.cls
```diff
@@ -2832,4 +2843,58 @@ ClassMethod SetConfiguredRemote(url) As %String
...
+        &sql(INSERT OR UPDATE INTO %SYS_Portal.Users (Username, Page, Data) VALUES (:username, :caption, :link))
+        
+        // Insert Git Pull link
+        set caption = "Git Pull: " _ installNamespace
+        set link = "/isc/studio/usertemplates/gitsourcecontrol/pull.csp?$NAMESPACE=" _ installNamespace
+        &sql(INSERT OR UPDATE INTO %SYS_Portal.Users (Username, Page, Data) VALUES (:username, :caption, :link))
+    }
+
+    // Update Security.Applications
+    &sql(UPDATE Security.Applications SET GroupById = '%ISCMgtPortal' WHERE ID = '/isc/studio/usertemplates')
```
isc-tleavitt, line 2867: This will fail as a no-op outside %SYS - probably should remove

csp/pull.csp
```diff
@@ -9,6 +9,18 @@
 <link rel="stylesheet" type="text/css" href="css/git-webui.css" />
 </head>
 <body>
+<pre for="contextSelect" style="padding-top: 2em;white-space: pre-wrap;height: 60px;">Change Context:  <select id="newContext" onchange="updateContext()">
+	<server>
+		set $NAMESPACE = %request.Data("$NAMESPACE",1)
```
isc-tleavitt, line 14: Should only change namespace if %request.Data("$NAMESPACE") is defined (just in case people have the old links bookmarked still, this will break badly)

isc-tleavitt (changes requested):
Two minor changes, otherwise good to go.

csp/pull.csp
```diff
@@ -11,12 +11,14 @@
 <body>
 <pre for="contextSelect" style="padding-top: 2em;white-space: pre-wrap;height: 60px;">Change Context:  <select id="newContext" onchange="updateContext()">
 	<server>
-		set $NAMESPACE = %request.Data("$NAMESPACE",1)
-		set contextList = ##class(SourceControl.Git.Utils).GetContexts(1)
-        set iterator = contextList.%GetIterator()
-		&html<<option>Select</option>>
-        while iterator.%GetNext(.key, .value) {
-				&html<<option value=#(value)#>#(value)#</option>>
+		if (%request.Data("$NAMESPACE")) {
```
isc-tleavitt, line 14: Now this could hit `<UNDEFINED>`. You need:
```
if $Data(%request.Data("$NAMESPACE",1),namespace)#2 {
set $Namespace = namespace
...
}
```

Bonus: make the current namespace selected in the `<select>`

csp/pull.csp
```diff
@@ -11,12 +11,14 @@
...
+			set $NAMESPACE = %request.Data("$NAMESPACE",1)
+			set contextList = ##class(SourceControl.Git.Utils).GetContexts(1)
+			set iterator = contextList.%GetIterator()
+			&html<<option>Select</option>>
+			while iterator.%GetNext(.key, .value) {
+					&html<<option value=#(value)#>#(value)#</option>>
```
isc-tleavitt, line 20: While we're at it, should put value attribute's value in double quotes for valid HTML

isc-tleavitt (changes requested):
Almost...

isc-tleavitt (approved):

isc-tleavitt:
Clearing review status - I want to play with this in HCC before giving the green light.

isc-tleavitt (approved):
Tested this out in HCC and it's just beautiful. Merging.

Thank you for your great work on this @isc-akadoo and also thanks to @isc-etamarch for putting up with my last minute nitpicks.

Issue #32, opened 2021-10-26
Document recommended use patterns


isc-tleavitt:
Specifically, should cover:

## Conceptual:
single vs. multi-developer (including namespace segmentation for shared environments as is commonly used in HCC)
remote vs. local instance and interaction with local git command line / GUI clients re: synchronization with IRIS

## Installation/Setup:
Initial setup - new repository
Initial setup - linking to existing repository

## Recommended workflow by persona:
"Hi, I'm an Interface Engineer. What's Git?": Hybrid approach with per-developer namespace, basic mode, main DEV namespace, deployment manager DEV->TEST and TEST->LIVE
"I'm an interface engineer, but this isn't my first CICD rodeo": Trunk-based strategy (with some coverage/material on GitLab and workflows)
Software developer: adapting to different workflows (e.g. [gitflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow)), advanced use with package manager, local dev / containers

# Priority list:
## HCC use cases:
- [x] Development workflow for Health Connect Cloud: basic mode, feature branch, merge request to development, git pull in shared development namespace (all via management portal)
- [ ] Initial setup for Health Connect Cloud: mapping configuration and "Import All (Force)" (all via management portal)
- [ ] Trunk-based strategy and CICD flow for Health Connect Cloud (changes get to dev/test/live via pipeline)

## Developer use cases:
- [ ] Use with standard branching/merging strategy
- [ ] Use with VSCode
- [ ] Use with package manager
- [ ] Use with local dev environment / containers

## Conceptual topics:
- [x] Conceptual introduction to source control / change control
- [ ] Conceptual: single developer vs. multi-developer
- [ ] Conceptual: remote vs. local instance and interaction with local git command line / GUI clients re: synchronization with IRIS

isc-svelury:
Also include a start-up guide and a high-level usage and feature guide.

isc-tleavitt:
Should also cover basic mode for the common Health Connect use case

PR #543, opened 2024-10-15
Add description of HCC workflow

isc-etamarch:
Addresses #32


isc-etamarch:
@isc-tleavitt Not sure if I got the merge request part of the HCC workflow correct.

docs/hcc.md
```diff
@@ -0,0 +1,48 @@
+# git-source-control for Health Connect Cloud
+
+## Development Workflow
+
+git-source-control has two main use modes: Basic and Expert. For Health Connect Cloud, we recommend beginning with basic mode, since it includes all of the necessary functionality for proper source control development while being easy to use for those with little development / source control experience. The recommended workflow for HCC is to create featyre branches, merge them into development, and then git pull the changes in the shared development namespace. The rest of this section explains this workflow more in depth for those with less source control experience. 
+
+### Feature branches
```
isc-tleavitt, line 7: We came up with the term "Interface branch" in a different discussion you may or may not have been part of; let's use that.

docs/hcc.md
```diff
@@ -0,0 +1,48 @@
...
+![Creating a new via sidebar](images/hcc/newbranch.png)
+![Naming new branch](images/hcc/newbranchnaming.png)
+
+Finally, you should click on the new branch you just made in the sidebar, and then press "Push".
+
+![Push your new branch](images/hcc/pushbranch.png)
+
+This makes sure that your branch will be tracked in Gitlab, so that other users can checkout your branch to test your changes on their own namespaces.
+
+After this, you can start working on the change.
```
isc-tleavitt, line 26: This is more of an expert mode workflow, really. I'm realizing that you might not have actually had enough context on basic mode to talk through the workflow. None of this is wrong, it's just too complicated for what we want to start with for someone with zero git background.

I'd honestly be fine with repurposing all you have as the expert mode documentation / as part of the coverage of Git WebUI. This is largely useful content still, just not for the bare-bones simplest-possible Health Connect Cloud use case.

isc-tleavitt (changes requested):
A few initial points of feedback - this is all accurate for more of an expert mode / Git WebUI-based workflow, but not actually the starting point for the Health Connect Cloud workflow.

isc-tleavitt (approved):
Just going to merge these because they're an incremental improvement; we can keep iterating.

PR #554, opened 2024-10-21
Added settings success banner

### LINKED ISSUE #550: Feedback after save
Add some sort of feedback at the top of the settings page after a save to show the user that the page was successfully saved

isc-etamarch:
Since saving causes us to move up to the top of the page, added a success banner at the top which will be visible

fixes #550

### DIFF
--- a/csp/gitprojectsettings.csp
+++ b/csp/gitprojectsettings.csp
@@ -68,6 +68,27 @@ body {
     padding-left: 4px;
 }
 
+.alert {
+  padding: 20px;
+  background-color: #04AA6D;;
+  color: white;
+}
+
+.closebtn {
+  margin-left: 15px;
+  color: white;
+  font-weight: bold;
+  float: right;
+  font-size: 22px;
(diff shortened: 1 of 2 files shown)


csp/gitprojectsettings.csp
```diff
@@ -134,7 +155,13 @@ body {
     do settings.%Save()
 }
 </server>
-<div class = 'container'>    
+<div class = 'container'>
+    <csp:if condition='$D(%request.Data("gitsettings",1)) && (##class(SourceControl.Git.Utils).NeedSettings() = 0)'>
+        <div class = "alert" style="display: 'none'">
```
isc-tleavitt, line 160: Why does this have display: 'none'? Note, I don't think this is valid anyway, should be display: none (but I think you really just don't want it to be there at all)

isc-tleavitt (changes requested):

isc-tleavitt (approved):

Issue #539, opened 2024-10-11
Fail more descriptively if you try to switch branches with conflicting uncommitted work


isc-tleavitt:
Right now, in basic mode, if I try to switch branches and it fails with something like:

```
error: Your local changes to the following files would be overwritten by checkout:
src/HCC/Foo/LearningServicesDemo.cls
Please commit your changes or stash them before you switch branches.
Aborting
```

This just gets buried in source control message output. We should fail more helpfully (e.g., with an alert) instead and make it clear to the user that the branch hasn't actually changed.

PR #557, opened 2024-10-21
Fixed checking out non-existent branch

### LINKED ISSUE #549: Switching to a branch that doesn't exist from menu breaks management portal
If you type-in a branch that does not exist into the "Checkout existing branch" , the management portal stalls and eventually stops responding. Users should get some sort of feedback that the checkout failed because the branch does not exist

isc-etamarch:
Fixes #549, #539

@isc-tleavitt Right now this throws an error (so that user is aware that checkout failed). Is there a better way to show an alert on the management portal / vscode / studio? (This happens on "AfterUserAction" so we have no access to the "Target" functionality of %Studio.SourceControl.Base.

### DIFF
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -327,7 +327,10 @@ ClassMethod AfterUserAction(Type As %Integer, Name As %String, InternalName As %
         }
     } elseif (menuItemName = "SwitchBranch") {
         if (Answer = 1) {
-            do ..SwitchBranch(Msg)
+            set status = ..SwitchBranch(Msg)
+            if $$$ISERR(status) {
+                Quit status
+            }
             set Reload = 1
         }
     } elseif (menuItemName = "Sync") {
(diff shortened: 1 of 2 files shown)


cls/SourceControl/Git/Utils.cls
```diff
@@ -327,7 +327,10 @@ ClassMethod AfterUserAction(Type As %Integer, Name As %String, InternalName As %
         }
     } elseif (menuItemName = "SwitchBranch") {
         if (Answer = 1) {
-            do ..SwitchBranch(Msg)
+            set status = ..SwitchBranch(Msg)
+            if $$$ISERR(status) {
+                Quit status
```
isc-tleavitt, line 332: Quitting with an error %Status is the right approach if something goes wrong in AfterUserAction; short of changing to a CSP page like sync I don't see a way to get a better UX than with this.

cls/SourceControl/Git/Utils.cls
```diff
@@ -400,6 +403,13 @@ ClassMethod NewBranch(newBranchName As %String) As %Status
 
 ClassMethod SwitchBranch(targetBranchName As %String) As %Status
 {
+    // Make sure that branch exists first
```
isc-tleavitt, line 406: What is current behavior (prior to this change) if you pick a branch that already exists on the remote but isn't local yet? What is the behavior after this change?

isc-tleavitt (changes requested):

cls/SourceControl/Git/Utils.cls
isc-etamarch: Now both should succeed

isc-etamarch:
I linked #539 since it was related and a quick addition to the current change. Now behavior is that checking out remote or local branches succeed unless they don't exist or you have uncommitted changes that will be deleted.

isc-tleavitt (approved):

PR #559, opened 2024-10-21
Add alert when other users have unstaged changes

### LINKED ISSUE #493: Warning on Sync page if other users have changes
Probably a low priority because basic mode is intended for single-developer namespaces, so ideally other users will not have any active changes.

If other users have changes in the namespace, the sync will fail to merge the remote trunk branch because there are unstaged changes in the repository. It's a little hard to tell what's going on from the output. We should add a warning message to the sync page that says that another user has unstaged changes, so merging the remote branch will fail.

isc-etamarch:
Fixes #493

### DIFF
--- a/csp/sync.csp
+++ b/csp/sync.csp
@@ -110,11 +110,14 @@
             <div class="row">
                 <div class="offset-sm-2 col-sm-8">
                     <server>
-                        if ##class(SourceControl.Git.Utils).InDefaultBranchBasicMode() {
+                        if ##class(SourceControl.Git.Utils).InDefaultBranchBasicMode() || ##class(SourceControl.Git.Change).GetOtherDeveloperChanges().%Size() > 0 {
+                            set message = $case(##class(SourceControl.Git.Utils).InDefaultBranchBasicMode(),
+                                1:"Please change branches to make sure your changes are committed.",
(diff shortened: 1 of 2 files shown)


csp/sync.csp
```diff
@@ -109,12 +109,18 @@
...
                 <div class="offset-sm-2 col-sm-8">
-                    <server>
-                        if ##class(SourceControl.Git.Utils).InDefaultBranchBasicMode() {
+                    <server> 
+                    set defaultBasic = ##class(SourceControl.Git.Utils).InDefaultBranchBasicMode()
+                        if defaultBasic || ##class(SourceControl.Git.Change).GetOtherDeveloperChanges().%Size() > 0 {
+                        if defaultBasic { 
+                            set message = "Please change branches to make sure your changes are committed." 
+                        } else {
+                            set message = "Other users have unstaged changes in this namespace, so merging to remote will fail"
```
isc-tleavitt, line 118: Nit: "unstaged" may be technically true but will confuse nontechnical users. Maybe a better phrasing would be:
"Other users have work in progress in this namespace. They must commit their work before you can sync, or you can commit their work after review in the Git WebUI before trying to sync again."

isc-tleavitt (changes requested):

PR #563, opened 2024-10-22
fix: production config page no longer closes UIs

### LINKED ISSUE #542: Prevent Sync / WebUI from closing when launched from Production Configuration page and reload is triggered in background due to timer
Right now this is very jarring and a pain in the neck.

There's a related upstream issue (internal ref: DP-424937) but I think git-source-control could do some clever things around this with either (a) a lock on the CSP session (which is a blunt instrument and possibly problematic) or (b) communication back to the parent frame.

isc-tleavitt:
This is a nice simple approach and it's working great in my testing locally.

Fixes #542

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added feedback to settings page (#550)
 - Fix "Home" navigation to point to current namespace (#548)
 - Force export of item if it has been modified (#354)
+- Production configuration page no longer closes Sync/WebUI when operations there change the production (#542)
 
 ## [2.6.0] - 2024-10-07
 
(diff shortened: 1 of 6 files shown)


isc-etamarch (changes requested):
I tested this out on HCC, and closing the UI after checkout does not cause a reload of production. Is the user supposed to reload themselves or do we expect the reload to occur by itself?

isc-tleavitt:
@isc-etamarch I observed the production reloading when I tried to do something on it, rather than via timer. I'll double check in HCC. In case it's relevant, which browser were you using?

isc-etamarch:
> @isc-etamarch I observed the production reloading when I tried to do something on it, rather than via timer. I'll double check in HCC. In case it's relevant, which browser were you using?

I am on edge. It reloads after an attempted change, but for changes such as actor pool size it allows you to save the changes before the reload happens (change persists after reload). Have not tested out making a change to non-existent / removed item yet

isc-etamarch (approved):
Very smooth transition!

Issue #342, opened 2024-05-31
Report authentication errors when pulling from SSH remote


isc-pbarton:
If I try to pull from an SSH remote with `##class(SourceControl.Git.API).Pull()` and my SSH key is not accepted, there are no errors reported. The return status is ok and the output looks like:
```
Pulling from branch: main
Fetch done
Files that will be modified by git pull:
    None
```
We should check for errors in the git fetch and output them to the user.

isc-pbarton:
I'm not able to replicate this anymore. I'm sure it's been addressed by other changes this year.

PR #565, opened 2024-10-22
Terminate git pull on error

### LINKED ISSUE #545: Terminate Git Pull if Errors
We should add argumnets to SourceControl.Git.API.Pull() method that should terminate a pull process with an OS level error if there is an error with the git pull, default to zero.

isc-etamarch:
fixes #545


isc-etamarch:
Should also fix #342 if the pull is done using the terminateOnError argument

isc-tleavitt (changes requested):
Ultimately these are two distinct issues.

The terminate from #545 is a reference to $System.Process.Terminate, as used here (for example): https://github.com/intersystems/ipm/blob/master/src/%25ZPM/PackageManager.cls#L432 - this is for use from CI/CD scripts/pipelines/etc.

We *don't* want to use this in a CSP context, ever. Terminating the process with an OS-level error is too heavy-handed. From other contexts, there is a possible improvement to error reporting/logging, but it's possible that other changes over the summer have made the original requirement/issue moot. Try testing out with a bad SSH key on an SSH remote and see what pull does now.

isc-etamarch:
Pravin tested and closed #342 because it no longer applies.

isc-tleavitt (changes requested):
Last thing - can you explicitly test that terminateOnError has the desired effect if an error occurs in the pull event handler? e.g., something fails to compile

isc-tleavitt:
> Last thing - can you explicitly test that terminateOnError has the desired effect if an error occurs in the pull event handler? e.g., something fails to compile

@isc-etamarch I can answer this through code inspection:
When SourceControl.Git.Utils:RunGitCommandWithInput runs the pull event handler, if an error occurs it puts the pull event handler output in errStream rather than outStream but doesn't return an error code.

We should have it return an error code in this case.

isc-etamarch:
> > Last thing - can you explicitly test that terminateOnError has the desired effect if an error occurs in the pull event handler? e.g., something fails to compile
> 
> @isc-etamarch I can answer this through code inspection: When SourceControl.Git.Utils:RunGitCommandWithInput runs the pull event handler, if an error occurs it puts the pull event handler output in errStream rather than outStream but doesn't return an error code.
> 
> We should have it return an error code in this case. Note, this is a higher risk change and we'll need to ensure other callers do the right thing. Alternatively we could do something hacky with a % variable we intentionally leak.

SourceControl.Git.Utils:Pull checks the output of errStream for error / fatal. Should we also check for an ERROR, and would that grab the issue?

isc-tleavitt:
> SourceControl.Git.Utils:Pull checks the output of errStream for error / fatal. Should we also check for an ERROR, and would that grab the issue?

Yes, that would do it.

isc-tleavitt:
@isc-etamarch on this: just check for ERROR and call it a day.

PR #566, opened 2024-10-22
Fix GitUI branch regex

### LINKED ISSUE #523: Branch with special characters not visible
Local branches that include special characters like . or @ (which are valid git characters), do not appear in the GitUI.

On creation of a branch (ex. test1.2.3) through the UI, the branch is checked out, but not shown in UI

isc-etamarch:
fixes #523


isc-tleavitt (approved):

isc-tleavitt:
@isc-etamarch I'm going to need to bring the cowbell back out for any time someone says "GitUI"

PR #589, opened 2024-10-28
Fix CSP imports

### LINKED ISSUE #251: CSP Mapping not working when pulling / importing files.
CSP files are not mapped/imported correctly. CSP files are exported and pushed to the Git repository correctly, but the other way around doesn't work. When pulling / importing files the are copied to the wrong place. 

Possibly the issue might have to do with the correct use of slashes. I tried replacing '/' with '\' in the mapping and also a mixture of both, But I couldn't figure out a working combination. Mostly to the files get imported to the csp root folder of the namespace.

isc-etamarch:
Fix #251 

Issue was ImportCSPFile attempting to delete a file that did not exist


isc-tleavitt (approved):
This makes perfect sense.

PR #594, opened 2024-10-29
Allow favorites changing for all users

### LINKED ISSUE #587: Favorites management - escalate roles so that user unable to manage own git favorites doesn't get stuck
Users with limited privileges can't add/edit their own favorites. The upstream issue for this (DP-435671) has been closed with no plans to fix it, so we need to work around the issue in git-source-control itself.

Proposed solution:
Use $$$AddAllRoleTemporary in a new %zpkg.isc.sc.git class to wrap calls to add favorites.

isc-etamarch:
Fixes #587 

@isc-tleavitt Are there any specific roles I should test besides %Developer?


cls/_zpkg/isc/sc/git/Favorites.cls
```diff
@@ -0,0 +1,55 @@
+Class %zpkg.isc.sc.git.Favorites Extends %RegisteredObject
+{
+    ClassMethod ConfigureFavoriteNamespaces(username As %String, newNamespaces As %String)
```
isc-tleavitt, line 3: newNamespaces is a %Library.DynamicArray

cls/_zpkg/isc/sc/git/Favorites.cls
```diff
@@ -0,0 +1,55 @@
+Class %zpkg.isc.sc.git.Favorites Extends %RegisteredObject
```
isc-tleavitt, line 1: No need to extend %RegisteredObject (though you might need to add include files if you drop it)

isc-tleavitt (changes requested):
I'd recommend restructuring as try/catch wrappers around calls to private methods. The try/catch wrappers should either validate the types of object arguments or better yet convert the objects to e.g. $ListBuild lists so that an attacker can't provide their own %GetNext/%GetIterator/etc. implementation that runs in a privileged context.

cls/_zpkg/isc/sc/git/Favorites.cls
```diff
@@ -0,0 +1,86 @@
+Class %zpkg.isc.sc.git.Favorites
+{
+    ClassMethod ConfigureFavoriteNamespaces(username As %String, newNamespaces As %Library.DynamicObject)
+{
+    // Convert to $listbuild
+    set namespaces = $lb("")
```
isc-tleavitt, line 6: Do you intend to have an empty option? An empty $listbuild list is just the empty string.

cls/_zpkg/isc/sc/git/Favorites.cls
```diff
@@ -0,0 +1,86 @@
...
+    for i=0:1:newNamespaces.%Size() {
```
isc-tleavitt, line 7: Would be slightly cleaner/probably faster to use the %GetIterator approach here

cls/_zpkg/isc/sc/git/Favorites.cls
```diff
@@ -0,0 +1,86 @@
...
+        &sql(CLOSE FavCursor)
+
+        if ('foundFlag) {
+            do nonFavNamespaces.%Push(namespace)
+        }
+    }
+    return {"Favorites": (favNamespaces), "NonFavorites": (nonFavNamespaces)}
+}
+
+ClassMethod SetFavs(username As %String, namespaces As %List) [Private] {
```
isc-tleavitt, line 65: Further hardening: make these NotInheritable too

isc-tleavitt (changes requested):
A few more comments.

In general, could you re-run a zpm "git-source-control activate" after making these changes so that we get proper UDL formatting? (e.g., around method definition lines with [Private] -> [ Private ] and the opening curly brace on a new line; if you don't do that it'll show up in someone else's future changes which is annoying)

isc-etamarch:
> A few more comments.
> 
> In general, could you re-run a zpm "git-source-control activate" after making these changes so that we get proper UDL formatting? (e.g., around method definition lines with [Private] -> [ Private ] and the opening curly brace on a new line; if you don't do that it'll show up in someone else's future changes which is annoying)

I have been. Maybe something in my local settings.

isc-tleavitt:
@isc-etamarch do you have git-source-control enabled + configured in the namespace you're operating against?

isc-etamarch:
@isc-tleavitt Ah that is probably the issue

isc-tleavitt (approved):

PR #608, opened 2024-10-30
Fix diffview

### LINKED ISSUE #590: Options at top of Git UI diff view (Ignore Whitespace, Complete file, etc.) should take effect immediately
These options can be quite useful but they are currently hard to use as they don't take effect immediately. If you click on them nothing happens but then if you click away to a different commit or file they do take effect.

image

IRIS for UNIX (Ubuntu Server LTS for x86-64 Containers) 2024.1 (Build 267_2U) Tue Apr 30 2024 16:04:37 EDT

isc-etamarch:
Fixes #590

### DIFF
--- a/git-webui/release/share/git-webui/webui/js/git-webui.js
+++ b/git-webui/release/share/git-webui/webui/js/git-webui.js
@@ -1524,14 +1524,8 @@ webui.DiffView = function(sideBySide, hunkSelectionAllowed, parent, stashedCommi
 
     var self = this;
 
-    self.update = function(cmd, diffOpts, file, mode, reRun) {
-        if (reRun) {
-            // Rerun with previous inputs
-            cmd = this.cmd
-            diffOpts = this.diffOpts
-            file = this.file
-            mode = this.mode
-        } else if (cmd || diffOpts || file || mode) {
+    self.update = function(cmd, diffOpts, file, mode) {
+        if (cmd || diffOpts || file || mode) {
             // if new input, update all
             this.cmd = cmd
             this.diffOpts = diffOpts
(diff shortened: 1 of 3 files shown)


git-webui/release/share/git-webui/webui/js/git-webui.js
```diff
@@ -1789,24 +1802,24 @@ webui.DiffView = function(sideBySide, hunkSelectionAllowed, parent, stashedCommi
 
     self.addContext = function() {
         self.context += 3;
-        self.update();
+        self.update(...[,,,,],1);
```
isc-tleavitt, line 1805: This is definitely going to break in Studio

git-webui/src/share/git-webui/webui/js/git-webui.js
```diff
@@ -1524,7 +1524,20 @@ webui.DiffView = function(sideBySide, hunkSelectionAllowed, parent, stashedCommi
 
     var self = this;
 
-    self.update = function(cmd, diffOpts, file, mode) {
+    self.update = function(cmd, diffOpts, file, mode, reRun) {
```
isc-tleavitt, line 1527: Cleaner approach: write a function that does everything after setting the 4 options, and call that instead of calling update with reRun set to true?

isc-tleavitt (changes requested):

isc-tleavitt (changes requested):
You need to re-run `zpm "git-source-control activate"` to get changes moved to release. (And then test, because if they're not in release you haven't tried them yet!)

isc-etamarch:
@isc-tleavitt should be all set now

isc-tleavitt (approved):

PR #620, opened 2024-11-01
Fix deleting files on import

### LINKED ISSUE #618: Deleting non-existent files still fails on import all

isc-etamarch:
Fixes #618


isc-pbarton (approved):
This doesn't need a CHANGELOG entry because it fixes a bug on an unreleased version, but looks good otherwise.

isc-tleavitt (approved):

PR #625, opened 2024-11-06
Errors switching to new branch are reported to UI

### LINKED ISSUE #624: Missing error handling when creating a new branch in basic mode
If I create a new branch in Basic Mode, and it fails to check out the default merge branch (for example because I have some uncommitted changes that conflict), it appears to succeed. A new branch is created off of my current branch instead of the default branch.

Then when I create a merge request to the default branch on my remote, my merge includes both my commits from the old branch and the new branch.

If we fail to checkout the default merge branch, we should display an error instead of creating a new branch.

isc-pbarton:
Fixes #624 which was experienced on a customer call today

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.8.0]
+## [2.7.1] - Unreleased
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.8.0]
```
isc-tleavitt, line 8: Again, should be 2.7.1 for now and need module.xml update

isc-tleavitt (changes requested):

isc-tleavitt (approved):

PR #628, opened 2024-11-07
Fix Unicode encoding of Git commands from web UI

### LINKED ISSUE #627: International characters in commit messages improperly encoded
The commit message in the Web UI no longer supports non-ASCII characters. If I input non-ASCII characters, the commit does succeed, but the commit message is mangled by being decoded to ASCII before the git command is issued. This is an old regression probably caused by #439.
Reported based on a user's discussion here: https://github.com/intersystems/git-source-control/discussions/621

isc-pbarton:
Fixes #627

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.8.0] - Unreleased
+## [2.7.1] - Unreleased
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.8.0] - Unreleased
```
isc-tleavitt, line 8: Need to update module.xml as well, and should be 2.7.1 for now in case we do a bug fix release

isc-tleavitt (changes requested):

isc-tleavitt (approved):

PR #634, opened 2024-11-08
add system settings

### LINKED ISSUE #535: Support system-wide default settings
Allow setting some global that will define default settings for when a new namespace is configured for git-source-control. This should include the default merge branch and basic mode. It should not include the mappings. The settings global will need to be accessible to users without %All permissions.

isc-etamarch:
Fixes #535 

@isc-tleavitt Let me know if settings should be added / removed from the ones saved


csp/gitprojectsettings.csp
```diff
@@ -633,7 +637,12 @@ body {
 
         <div class="form-group row mb-3">
             <div class="col-sm-12 text-center">
-                <input type='submit' class="btn-lg btn-primary" value = 'Save'/>
+                <input type="hidden" id="proxySubmitButton" name="proxySubmitButton" value="Save">
+                <input type='submit' class="btn-lg btn-primary" value = 'Save' name="Save"/>
+                <input type="submit" class="btn-lg btn-primary" id="saveDefaults" value = 'Save as Instance Defaults' title="Save these settings as default for new instances" name="saveDefaults"/>
+                <div>
+                    <small> Settings not saved as default marked with '*'</small>
```
isc-tleavitt, line 644: This generally means "required" which I think is confusing UX.

cls/SourceControl/Git/Settings.cls
```diff
@@ -394,4 +395,25 @@ Method ConfigureBinPath(ByRef path As %String) As %Boolean
...
+    while iterator.%GetNext(.key, .value) {
+        set $property($this,key) = value
+    }
+    return $$$OK
+}
+
+Method SaveDefaults() As %Boolean
+{
+    set defaults = {}
+    set items = $lb("gitBinPath", "namespaceTemp", "pullEventClass", "percentClassReplace", "environmentName", "systemBasicMode", "defaultMergeBranch", "mappedItemsReadOnly", "compileOnImport")
```
isc-tleavitt, line 411: namespaceTemp shouldn't be a system-level default! This will always differ per namespace as things stand. Even once we add support for instance-wide, this is going to be more complex to configure (probably).

csp/gitprojectsettings.csp
isc-tleavitt: Recommendation: put labels for settings that *are* saved as defaults in bold.

isc-tleavitt (changes requested):
Note: from a SemVer perspective, "Added" means we're going from a patch (bugfix) release to a minor release, so the changelog and module.xml should update to 2.8.0 accordingly.

isc-tleavitt (changes requested):
One more nit, let's put Added before Fixed in the changelog, should have said that before.

I'm going to hold this in Approved until we decide if we're going to do a bugfix release prior to releasing Production Decomposition.

isc-tleavitt (approved):

Issue #531, opened 2024-10-08
Popping from stash (with merge conflict) causes issues


isc-etamarch:
Based on customer feedback

If popping changes from stash results in a merge conflict, the GitUI errors out, and the production is left in a broken state (in merge conflict state). There is no way to fix this through the UI, only manually fixing the merge conflict in the code. 

We need to allow for a way to resolve the merge conflict without going directly into the code.

@isc-tleavitt What are your thoughts?

isc-tleavitt:
I think the right option would be:
Attempt production configuration auto-resolve
If that fails, cancel the stash pop/apply (ideally preserving the stash?)

This is also a pointer toward recommending basic mode / per-developer namespaces to the customer.

isc-etamarch:
This also causes the workspace view to error out. Trying to commit this change does not work, though discarding this file does not actually discard the file
![image](https://github.com/user-attachments/assets/90baea79-d716-4755-89a3-3ecb54256166)

isc-tleavitt:
The other resolution option is to use the Web Application view of the repository, which gives you isfs access to the raw files. But we can't have that as the only/default option.

PR #653, opened 2024-11-21
Popping from stash

isc-etamarch:
Fixes issue #531

### DIFF
--- a/cls/SourceControl/Git/WebUIDriver.cls
+++ b/cls/SourceControl/Git/WebUIDriver.cls
@@ -218,27 +218,61 @@ ClassMethod HandleRequest(pagePath As %String, InternalName As %String = "", Out
                 } else {
                     set inFile = ""
                 }
+                // Want to invoke merge conflict autoresolver in case of issues
+                if (($piece(args, " ", 1, 2)  = "stash pop") || ($piece(args, " ", 1, 2)  = "stash apply")) {
+                    set succeeded = 0
+                    set initTLevel = $TLEVEL
+                    try {
+                        TSTART
+                        set returnCode = ##class(SourceControl.Git.Utils).RunGitCommandWithInput("-c",inFile,.errStream,.outStream,gitArgs...)
+                        if (returnCode '= 0) {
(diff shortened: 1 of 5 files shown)


cls/SourceControl/Git/WebUIDriver.cls
```diff
@@ -218,27 +218,61 @@ ClassMethod HandleRequest(pagePath As %String, InternalName As %String = "", Out
                 } else {
                     set inFile = ""
                 }
+                // Want to invoke merge conflict autoresolver in case of issues
+                if (($piece(args, " ", 1, 2)  = "stash pop") || ($piece(args, " ", 1, 2)  = "stash apply")) {
```
isc-tleavitt, line 222: Over time we should be moving from /git to /git-command. Wondering about whether it makes sense to make that move for stash now and put this handling there.

isc-tleavitt: Also wondering about whether we should take the extra time to solve the general problem and cover merge (via WebUI) as well.

isc-tleavitt: At minimum we should refactor "run git command and auto-resolve conflicts" to its own method used from everywhere it's relevant.

git-webui/src/share/git-webui/webui/js/git-webui.js
```diff
@@ -1895,8 +1895,14 @@ webui.DiffView = function(sideBySide, hunkSelectionAllowed, parent, stashedCommi
             return;
         }
         var stashIndex = parseInt($(".log-entry.active .stash-list-index").text());
-        webui.git("stash apply stash@{"+stashIndex+"}", function(output){
-            webui.showSuccess(output);
+        $.post("git", {command: "stash apply stash@{"+stashIndex+"}"}, function(output) {
+            if (output.substring(0,5) == "ERROR") {
```
isc-tleavitt, line 1899: Would rather keep cleaner API for communicating error status (using HTTP error codes / returnCode w/ standardized handling of an error response)

isc-tleavitt (changes requested):

isc-tleavitt (approved):

PR #666, opened 2024-12-11
Add export production option to menu for productions

### LINKED ISSUE #665: Add option to export production from source control menu
Currently if you are enabling or disabling production decomposition, there is a step to go into a terminal and run BaselineProductions to reexport productions to the repository in the new format. We should make this possible for users without access to an IRIS terminal.
Proposal: add a new option "Export Production" to the source control menu only if the currently opened item is an interoperability production. If a different user has the production class or any items checked out, it should fail with a descriptive error message. Otherwise it should export the production in the configured composed/decomposed format, and delete any items from the repository associated with the production in the other format.

isc-pbarton:
Fixes #665


isc-tleavitt (changes requested):
nitpick: version is 2.9.0 with something added, module.xml should also be updated if it hasn't been yet

isc-tleavitt (approved):

PR #683, opened 2025-01-13
API for loading decomposed productions

### LINKED ISSUE #670: Support continuous deployment of decomposed production items
We need an API method that supports production decomposition for continuous deployment workflows that deploy into environments that are not themselves configured with Embedded Git as a source control extension. This is important for HealthConnect Cloud deployment.
The API should take in a directory path with subdirectories that each contain PTD items for a different production. It should deploy all of these items to the appropriate production. It should also handle deleting hosts in the production that do not exist in the directory (not including built-in hosts like the testing and monitoring services).

isc-pbarton:
Resolves #670


isc-tleavitt (approved):
This looks good to me - need to decide what to do with 2.9.1 and its changes though. Is that just 2.10.0 or do we want a bugfix release? Thoughts?

isc-pbarton:
None of the bugs fixed in 2.9.1 were introduced in 2.9.0, so my inclination would be to roll those changes into 2.10.0

PR #686, opened 2025-01-14
Add Business Processess if they are not in source control

### LINKED ISSUE #676: Business processes are not automatically added to source control
Most items get automatically added to source control when you create them. Business processes on the other hand you have to remember to click the "Add to Source Control" option on. This is an easily forgotten step.

isc-etamarch:
Fixes #676


isc-etamarch:
@isc-pbarton This is only a partial solution. This works only for compiled business processes, since there is no easy way to check (that I know of) if a class is a business process before it has been compiled.

isc-pbarton:
I'm seeing it with business rules if I start from the production configuration page. Add a new process of type EnsLib.MsgRouter.RoutingEngine, check "Auto-create rule", then open the auto-created rule. It has the "Add to Source Control" option on it. This is on IRIS 2023.1.3

isc-etamarch:
@isc-pbarton That should work now too, though the Rule Editor does not call hooks when opened, so this will only work on Save/Compile

isc-tleavitt (approved):

PR #689, opened 2025-01-16
Fix Discard of deleted files

### LINKED ISSUE #688: Discard and Stash fail on 'delete' operation
If we deleted an item, and want to discard / stash this delete, we get an error that the file does not exist. Need to fix functionality for reverting / stashing delete changes

isc-etamarch:
Fixes #688

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed not showing warnings on Studio (#660)
 - Fixed business processes and rules not being added to source control automatically (#676)
 - Embedded Git commits settings when cloning empty repo to avert any issues
+- Fixed Discard / Stash not working on deletes (#688)
 
 ## [2.9.0] - 2025-01-09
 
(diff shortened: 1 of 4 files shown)


isc-etamarch:
I did not fix stash, since stashing deleted files is frustratingly complex. The best solution would be 'git stash push --staged' which would stash all staged files (just the deleted files), but this seems to break the unstaged files a little, where we now cannot see a diff from the Git UI, and git loses track of the type of operation (i.e. if a file is new, changed, etc.).

Until I find a solution that doesn't break everything, recommendation would be to discard deletions (especially since deleting is an easy operation to redo if needed).

isc-pbarton:
@isc-etamarch I think this fixes a somewhat different issue than what we saw today that motivated #688. I'm still able to replicate that one by:
- Add a production to source control and commit it.
- Use the "List Productions" page to delete the production
- In the Git repository, run "git add <file>" to add the delete to the index. (This simulates what might happen if I tried to commit the delete through Web UI, the add succeeds, but the commit fails).
- In the Git Web UI, select the deleted file and use the "Discard" option. It fails with "fatal: pathspec 'cls/Diagram/SampleProduction.cls' did not match any files" because it is trying to do a "git add" but the file was already removed from the index.

This goes through a different code path that doesn't involve SourceControl.Git.DiscardState

isc-etamarch:
@isc-pbarton I see. That last commit should fix that as well (The other changes I made are still necessary for it to work)

isc-tleavitt (approved):
One really minor nitpick - feel free to tack on a commit to fix, then merge.

Issue #586, opened 2024-10-28
Provide a way to undo commits in the Git WebUI


isc-tleavitt:
e.g., an equivalent to git reset --hard HEAD~1

PR #691, opened 2025-01-17
Added revert / reset Head

isc-etamarch:
Addresses #586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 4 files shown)


isc-etamarch:
@isc-tleavitt / @isc-pbarton Could you review the description text to make sure that it sounds correct. Also, should we have another confirmation dialog after the revert type has been chosen?

Description Text:
'There are a few options available to revert the previous commit. Please read the description carefully to make sure you choose the correct option.

Revert: git revert --no-commit - This will create a new change, which will be the reversal of the previous commit. 'It will not be automatically committed, so you can inspect/modify/combine the changes with others before you commit.

Hard Reset: git reset --hard HEAD~1 - This will delete the previous commit entirely, and reset you to a state before the commit. WARNING: This will also delete <b>all uncommitted changes</b>, so make sure you have no changes left before attempting this operation.'

isc-tleavitt (approved):

PR #695, opened 2025-01-24
Improvements to deployment of production items

### LINKED ISSUE #690: Better handling of timeout errors with deployment of decomposed productions
The following can occur in a large decomposed production with a lot of active development:

- I try to get the latest and greatest code by using the Pull option in the Source Control menu. It times out, eventually just giving me an alert that says "null".
- I pull changes through the terminal by running `do ##class(SourceControl.Git.API).Pull()`. It updates the local repository, but deployment fails because it claims there is an in progress deployment.
- I go into the deployment history in the Management Portal and clear the failed deployment.
- I deploy again by running `do ##class(SourceControl.Git.API).ImportAll()`. It takes a minute but finally succeeds.

isc-pbarton:
Resolves #690 

Prior to this change we were calling ImportItem() twice for every item changed by a Git operation. Once through AddToServerSideSourceControl, and once through the pull event handler. This made every Git operation changing productions twice as long as it needed to be. This was only an issue for production items because other items have a timestamp from %RoutineMgr that lets us skip the import if the item is up to date in IRIS.

Also adds a separate table that logs deployments into IRIS through the pull event handler.

### DIFF
--- a/cls/SourceControl/Git/PullEventHandler.cls
+++ b/cls/SourceControl/Git/PullEventHandler.cls
@@ -24,6 +24,10 @@ Method OnPull() As %Status [ Abstract ]
 /// <var>pullEventClass</var>: if defined, override the configured pull event class
 ClassMethod ForModifications(ByRef files, pullEventClass As %String) As %Status
 {
+    set log = ##class(SourceControl.Git.DeploymentLog).%New()
+    set log.HeadRevision = ##class(SourceControl.Git.Utils).GetCurrentRevision()
+    set log.StartTimestamp = $zdatetime($ztimestamp,3)
+    do log.%Save()
     set event = $classmethod(
         $select(
             $data(pullEventClass)#2: pullEventClass, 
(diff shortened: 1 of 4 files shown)


cls/SourceControl/Git/Utils.cls
```diff
@@ -764,10 +765,6 @@ ClassMethod AddToServerSideSourceControl(InternalName As %String) As %Status
             continue
         }
         set @..#Storage@("items", item) = ""
-        #dim sc as %Status =  ..ImportItem(item, 1)
```
isc-tleavitt, line 767: Review impact of this on setting timestamps

cls/SourceControl/Git/PullEventHandler.cls
```diff
@@ -24,14 +24,22 @@ Method OnPull() As %Status [ Abstract ]
 /// <var>pullEventClass</var>: if defined, override the configured pull event class
 ClassMethod ForModifications(ByRef files, pullEventClass As %String) As %Status
```
isc-tleavitt, line 25: Tighten with try/catch and double check exception handling of callers (make sure they're using the returned %Status)

isc-tleavitt (changes requested):

cls/SourceControl/Git/Utils.cls
isc-pbarton: I did some code review and testing - I don't think there's going to be any impact on timestamps. In the original diff that added this #125 the intent was to use this to deploy items added in the pull. Since then we've moved that logic back to the pull event handlers, so this is purely a duplicate. Timestamps will get updated because we call ImportItem() separately in every code path where we call AddToServerSideSourceControl().

isc-tleavitt (approved):

PR #707, opened 2025-02-14
Add IPM event handler that does an uninstall

### LINKED ISSUE #631: IPM event handler to do an uninstall/install
Use case: class deleted from IPM package in branch A
Switch to branch B which still has class - gets reloaded
Back to branch A - class is in IRIS and gets reexported on zpm "load"

isc-pbarton:
Resolves #631


isc-tleavitt (approved):
Great solution, I like it!

PR #735, opened 2025-04-08
Fixes #655

### LINKED ISSUE #655: SystemMode in CPF overrides environment name in Git settings
I can use the "environment name" control in the Git settings page to set the system mode for the instance (DEVELOPMENT,TEST,LIVE). However if there is a SystemMode defined in the iris.cpf file, that will override the one we set when the instance restarts. We'll need to use Config.Startup to set it directly in the CPF (or document that system mode in CPF will override the one we set here).

isc-etamarch:


cls/_zpkg/isc/sc/git/SystemMode.cls
```diff
@@ -0,0 +1,26 @@
+Class %zpkg.isc.sc.git.SystemMode
+{
+    ClassMethod SetEnvironment(environment As %String) As %Status [ NotInheritable, Private ]
+{
+    $$$AddAllRoleTemporary
+    do $SYSTEM.Version.SystemMode(environment)
+    set nm = $NAMESPACE
```
isc-pbarton, line 7: We should `new $namespace` before changing namespace so that the namespace change is scoped to this method.

cls/_zpkg/isc/sc/git/SystemMode.cls
```diff
@@ -0,0 +1,26 @@
...
+    zn "%SYS"
+    set obj = ##class(Config.Startup).Open()
+    set obj.SystemMode = environment
+    do obj.%Save()
```
isc-pbarton, line 11: Check status returned from %Save()

isc-pbarton (changes requested):
A couple small code changes requested.

isc-pbarton (approved):

PR #744, opened 2025-04-17
Compatibility with method signature changes in %Studio.SourceControl

### LINKED ISSUE #743: Installation fails with latest %Studio.SourceControl
Installing Embedded Git fails with a method signature compilation error with IRIS change EPG162 (not yet released).

isc-pbarton:
Fixes #743


isc-pbarton:
Not backwards compatible. We will have to either change upstream or remove the dependency in Embedded Git.

Issue #756, opened 2025-04-23
Different Mappings per namespace, get overwritten due to embedded-git-config.json


pisani:
Take an enterprise solution consisting of 2 interoperability production, running in two different namespaces., and one git repository, (called for argument's sake - "MySolution").  

The IRIS instance defines a git initialise root folder:  c:\MySolution.
The IRIS instance defines namespaces APPA and APPB where these productions are developed, and I want code edited/saved in each namespace to go to a namespace appropriate sub-folder.

The export mappings as therefore defined in Settings UI say, 
- for APPA, that CLS files export to /APPA/src/cls, INC to /APPA/src/inc, etc. 
- for APPB, that CLS files export to /APPB/src/cls, INC to /APPB/src/inc, etc.  - you get he idea...

This works fine - initially...., but changing these settings via the UI creates an export file:  embedded-git-config.json in the ROOT work folder, with the most recently updated namespace values (which could be APPA or APPB), and re-entering into the other namespace's Settings UI, loads and *overwrites* these previously defined namespace-specific mappings.

What is the use of this embedded-git-config.json ?  Am I forced to a one-to-one mapping of GIT Repository to Namespace ?

pisani:
I couldn't see how adding embedded-git-config.json to .gitignore allow me to manage the mapping configurations seperately for each namespace, but I tried this anyway (and it fails).

The process:
when you pull up the Settings (say for APPA), and make a small change and Save the settings - a default green  bar at the top of the settings UI windows says settings have been **exported** to embedded-git-config.json.
When you then switch the other namespace (eg APPB), pull up its settings and make some minor change, and Save - the green bar says:

" embedded-git-config.GSC has been **imported** from C:\MySolution\embedded-git-config.json exporting new version of embedded-git-config.GSC to C:\MySolution\embedded-git-config.json "

The first part is tell-tale clue to the issue.  It seems, on every SAVE, the settings are first imported from file.  This is unrelated to whether the file name is added to .gitignore or not.

pisani:
to generate a namespace-specific embedded-git-config.json file I made these changes to SourceControl.Git.Settings.Document:

```
Parameter INTERNALNAME = {"embedded-git-config-"_$namespace_".GSC"};
Parameter EXTERNALNAME = {"embedded-git-config-"_$namespace_".json"};
```

This embeds the namespace as part of the filename, and are kep seperate.   Is this a valid solution to the namespace-specific config file that is worthy of a pull request ?

[...]

PR #762, opened 2025-04-29
Mappings expand <namespace> parameter

### LINKED ISSUE #710: Support namespace scoping for git config file
We should permit having a different embedded-git-config.json on a per-namespace level in order to play nicer with production decomposition, as customers might configure separate local repositories for each namespace and pull into a single remote. Currently we only support having a global config file for the entire repo.

isc-pbarton:
Mapping configurations may now include the \<namespace\> token. The token will be expanded to the name of the current namespace. This is the first step towards supporting an instance-wide monorepository in which custom code for each namespace is mapped to a different subdirectory.
Resolves #710 and #756


isc-pbarton:
Making this into a draft - I will add some improvements here that allow isolated developer namespaces for a multi-namespace solution.

isc-etamarch (approved):

Issue #783, opened 2025-06-02
Checkout of remote branch does not sync files with IRIS


raymond-rebbeck:
On latest main branch (likely previous versions but have not tested) if you checkout a remote branch (one that does not exist as a local branch) then the files are not synced with IRIS.

This appears to be related to the below block of code where the diff for the sync is performed prior to the checkout command being performed. One cannot simply perform a diff on a remote branch without either already having checked it out or prefixing the name with 'origin/' or something similar. Therefore the diff sees no differences and nothing is synced.

https://github.com/intersystems/git-source-control/blob/4cea6ab7a1c07446673ac0d04596fef413011c7b/cls/SourceControl/Git/Utils.cls#L1970-L1979

My initial thoughts are that something similar to what was done for the pull command in #517 could be applied so that the diff is performed after the checkout command is run. It's slightly more complicated though as a checkout can be performed both via the IDE menu and from Git UI. Both situations would need to be handled.

IRIS for UNIX (Ubuntu Server LTS for x86-64 Containers) 2024.1 (Build 267_2U) Tue Apr 30 2024 16:04:37 EDT

PR #788, opened 2025-06-05
Sync IRIS with repo changes when switching to a remote branch

raymond-rebbeck:
Addresses #783 

First commit re-factors the 'diff after git command' mechanism previously added in #517 so that it can be more generically applied to other Git commands.

Second commit applies it to the checkout command.

Works with Git menu dropdown and with WebUI. The diff that was previously being applied prior to the checkout being performed would fail due to the branch not yet being local at that point in time when checking out a remote branch for the first time.


isc-pbarton (approved):
works great, thank you!

PR #800, opened 2025-06-27
Fix initial import of settings file that has yet to be imported

### LINKED ISSUE #794: Timestamp of embedded-git-config.GSC is always the current time if not yet imported, preventing import
If the config file has yet to be loaded (because for example I only just installed git-source-control and did nothing else) then the `%RoutineMgr.TS()` timestamp for `embedded-git-config.GSC` is always returned as the current time. This results in not being able to then subsequently import the config file as it is never considered outdated (at least without using force=1).

The following code appears to be the reason for this, if `^SYS("SourceControl","Git","settings","TS")` is not set then the current time is returned as the default `$get` value. The comments for this class method seem to be suggesting that the default should be `""`, with the referenced `$zdatetime($horolog,3)` being what would be in the global if it existed.

https://github.com/intersystems/git-source-control/blob/573c292c9a67bf9f52959c1af9a96a96a0e8e443/cls/SourceControl/Git/Settings/Document.cls#L100-L106

The reason for being in this situation in the first place is that I want to automate the setup of Embedded Git for end user developers. I already do this in 2.7.1 by setting a selection of `^SYS("SourceControl","Git","settings",...)` globals via a script, including mappings, and then executing `SourceControl.Git.Utils.ImportAll()` to import all of the code. It works great.

However as part of exploring a transition from 2.7.1 to 2.12+ I want to use the config file for the mappings and ran into this issue while trying to perform an initial import of the config file.

raymond-rebbeck:
Fixes #794 

Do not return the current time as a timestamp when the settings file has never been imported.

Returning the current time can prevent import of the settings file as it can prevent it from being considered outdated.

The comment of the timestamp message very much suggests to me that returning `""` is the intended behaviour and that `$zdatetime($horolog,3)` is just a description of the format of the timestamp that would be returned if there was one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed importing Lookup Tables that do not already exist (#791)
+- Fix initial import of settings file that has yet to be imported (#794)
 
 ## [2.12.1] - 2025-06-27
 
--- a/cls/SourceControl/Git/Settings/Document.cls
+++ b/cls/SourceControl/Git/Settings/Document.cls
@@ -102,7 +102,7 @@ Method UpdateHash(stream)
 /// or "" if the routine does not exist.
 ClassMethod TimeStamp(name As %String) As %TimeStamp
 {
-    return $get(@##class(SourceControl.Git.Utils).#Storage@("settings","TS"), $zdatetime($h,3))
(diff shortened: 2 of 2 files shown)


raymond-rebbeck:
It's strange that a test case failed. It did not happen when I initially created this PR and all I've done is rebase this branch on main to pick up the 2.12.1 release changes.

raymond-rebbeck:
I rebased again and the test case succeeded again. I had a look through the relevant code and it was not immediately apparent how this might have been responsible for the failure. Perhaps there is some kind of race condition that coincidentally happened.

PR #807, opened 2025-07-11
Import embedded-git-config.json first when performing an Import All

### LINKED ISSUE #806: Import All should load the config file before anything else
Similar to #697 which applies to pull event handlers. Import All should load `embedded-git-config.json` first to avoid issues where the mapping configuration has changed.

If a config file has never been loaded before such as after having just installed Embedded Git then it should be possible to do a single Import All where it imports the config file first and then proceeds to import everything else. Currently this is only possible with multiple invocations of Import All which is inconvenient.

raymond-rebbeck:
Fixes #806 

The config file `embedded-git-config.json` is now imported first when performing an Import All.
- It is only imported once
- If it is imported then the item list is obtained again in case there are mapping configuration differences after import

This combined with some other recently merged PRs (#792, #799, #800) allows for an existing Git repository that contains a config file to be imported with the only initial configuration required being the repository location: `^SYS("SourceControl","Git","settings","namespaceTemp")` (if your Git path and other defaults are acceptable at least)

### DIFF
--- a/cls/SourceControl/Git/Utils.cls
+++ b/cls/SourceControl/Git/Utils.cls
@@ -1567,7 +1567,31 @@ ClassMethod ImportRoutines(force As %Boolean = 0, pullEventClass As %String) As 
     
     #dim ec as %Status = ..ListItemsInFiles(.itemList, .err)
     quit:'ec ec
-
+    
+    // If there is a config file it must be imported before everything else.
+    set internalName = ""
+    for {
+        set internalName = $order(itemList(internalName))
+        quit:internalName=""
+        
+        // If a config file is found then import it
+        if internalName = ##class(SourceControl.Git.Settings.Document).#INTERNALNAME {
+            set sc = ##class(SourceControl.Git.Utils).ImportItem(internalName, force)
+            
+            if $$$ISERR(sc) {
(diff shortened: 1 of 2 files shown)


cls/SourceControl/Git/Utils.cls
```diff
@@ -1567,14 +1567,42 @@ ClassMethod ImportRoutines(force As %Boolean = 0, pullEventClass As %String) As
...
-
+    
+    // If there is a config file it must be imported before everything else.
+    set internalName = ""
+    for {
+        set internalName = $order(itemList(internalName))
+        quit:internalName=""
+        
+        // If a config file is found then import it
+        if internalName = ##class(SourceControl.Git.Settings.Document).#INTERNALNAME {
```
cambot, line 1578: For what it's worth, I think your for loop (lines 1572-1578) could be simplified to a single if statement:

```
if $Data(itemList(##class(SourceControl.Git.Settings.Document).#INTERNALNAME))
```

raymond-rebbeck: Thank you. I agree and have updated the branch to do this. I have added you as a co-author.

PR #846, opened 2025-08-28
Fix STACK errors in web UI

### LINKED ISSUE #845: FRAME error when discarding many changed files
This was witnessed on a HealthConnect instance after some unrelated gremlin changed the directory permissions of the git repository, causing every file to be shown as edited. We tried to select all through the Git Web UI and discard the changes, but gotten an "application error" popup. The Application Error Log showed a <FRAME> error.

The resolution was to go into the operating system and run git clean -df.

isc-pbarton:
Fixes #845

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed Import All deploying changes to files in a CSP application (#828)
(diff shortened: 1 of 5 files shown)


isc-pbarton:
Deferring this for now. The linked commit fixes the \<STACK\> error, but on Docker we run into another \<MAX $ZF STRING\> error when we try to make a call to $zf(-100) with more than a hundred arguments. A fix for the future will have to involve limiting or batching these operations from the web UI.

Issue #215, opened 2022-12-19
Support instance-wide monorepo


isc-tleavitt:
Configuration/design would tie in to #214 . WebUI should support staging/commits with a folder-based filter (current namespace only or cross-namespace for the cases where it matters).

isc-tleavitt:
Note: #214 was done independently of this to support some use cases (like HCC) where we want each namespace to be able to use Embedded Git independently.

isc-tleavitt:
From discussion with @isc-pbarton today - with an instance-wide monorepo:
* Embedded Git should support automatically configuring mappings for settings so that settings and uncommitted queue live instance-wide
* We need (as a prerequisite/corequisite) #710 to support intentionally-differing behavior between namespaces
* Like we have a token in configuration for environment, need the same for namespace.
* "Import All" and pull event handler needs to be able to span namespaces. To do this, we'll need a new instance-wide pull event handler, and we need to be able to parameterize it with the order in which to import.
* "Export All" + baseline likewise need to be able to span namespaces
* Uncommitted queue operations need to be able to backtrack to apply to the correct namespace(s) - this is really tricky but simplified because everything goes through the pull event handler.

So "Instance-wide enabled" means:
* Add mappings
* Set proper pull event handler
* Improper pull event handlers should complain/not be allowed - must be or extend our built-in instance-wide pull event handler

isc-tleavitt:
For protection of HCC (which won't want this... probably?), we should check to see if mappings already exist but not via %ALL (e.g., REPO) and descriptively refuse to enable instance-wide if they do.

PR #851, opened 2025-09-04
Document instance-wide monorepository

isc-pbarton:
Added some pointers to using Embedded Git to source control multiple namespaces with a single local repository.
Relevant to #215

### DIFF
--- a/docs/instance-wide.md
+++ b/docs/instance-wide.md
@@ -1,0 +1,107 @@
+# Instance-wide Git Configuration
+
+## Motivation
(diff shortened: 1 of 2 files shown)


isc-pbarton:
Here is a friendlier link to the draft document:
https://github.com/intersystems/git-source-control/blob/doc-map-instance-wide/docs/instance-wide.md
I pulled this together to match how our customers using CCR do instance-wide source control.

@cambot I'm especially interested in your thoughts here. The major difference from what we discussed before is that all application namespaces share a single routines database. This will be necessary if there is some shared code that is used in multiple namespaces.

isc-pbarton:
My local testing shows this setup works out. A few more updates to the documentation needed:
- You have to install Embedded Git after step 3
- You have to create a package mapping for %zpkg.isc.sc.git to the namespace Embedded Git is installed in. There's another GitHub issue to create that mapping in MapEverywhere. We should fix that, and document running MapEverywhere as one of the steps here.

isc-dchui (approved):
I like the diagram!

PR #860, opened 2025-09-17
Cloning an existing repo with Configure overwrites settings from file

### LINKED ISSUE #819: Configure script overwrites existing embeded-git-config.json file
I have noticed that when I run the `Configure()` script and choose the option to clone a repository, the `embeded-git-config.json` present in the clone gets overwritten by the `Configure()` script.  This means we need to take the extra step of discarding the change and running "Import All (force)" to ensure that our source code is loaded correctly into a target environment.

We have Mac and Linux environments that are pretty locked down, so this step turns into additional steps because the git checkout is owned by `irisusr` and our developers cannot simply run `git reset --hard HEAD`.

1. At a minimum, if "Clone Repository" is selected, I do not believe it should export a config file based on what was selected in the setup prompts.
2. If you ask this question earlier (and the answer is yes), I think you can simplify the configuration flow by skipping questions whose answers end up in the config file.

isc-pbarton:
resolves #819

If the cloned repo has a settings file, its values will overwrite those chosen earlier in the Configure prompts. I agree with the requestor that this is always desired behavior.

### DIFF
--- a/csp/gitprojectsettings.csp
+++ b/csp/gitprojectsettings.csp
@@ -119,6 +119,7 @@ body {
         }
 
         if ('settings.settingsUIReadOnly) {
+            do ##class(SourceControl.Git.Utils).Locked($get(%request.Data("lockNamespace",1)))
             for param="gitBinPath","namespaceTemp","privateKeyFile","pullEventClass","percentClassReplace", "defaultMergeBranch","environmentName","mappingsToken" {
                 set $Property(settings,param) = $Get(%request.Data(param,1))
             }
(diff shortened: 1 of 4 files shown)


isc-pbarton:
Includes changes from #859 as well because I did not feel like resolving a merge conflict later.

csp/gitprojectsettings.csp
isc-dchui, line 1: So this PR also contains the changes for #859. Is that intentionally to avoid a merge conflict later? If so, will the git history correctly reflect that these changes are not actually from this PR?

isc-dchui (commented):
Just a quick question.

isc-dchui (approved):

csp/gitprojectsettings.csp
isc-pbarton: @isc-dchui Yep, that was intentional to avoid merge conflict on the changelog. The history should be all good now - I merged #859 and changed the base on this repo to force GitHub to refresh its data associating commits to PRs.

PR #931, opened 2026-02-18
fix(IncrementalLoad): Do not attempt to delete files not in the datase

### LINKED ISSUE #928: <SUBSCRIPT> error in IncrementalLoad pull handler.
I was switching between branches when I noticed that the changed files were not making it into my Iris instance.  When I examined the DeploymentLog, I found this error:

```ObjectScript
<SUBSCRIPT>OnPull+21^SourceControl.Git.PullEventHandler.IncrementalLoad.1 *delList() Subscript 2 is ""
OnPull+21^SourceControl.Git.PullEventHandler.IncrementalLoad.1
```

This is that line of code:
```ObjectScript
if ((internalName = "") && (..ModifiedFiles(i).changeType '= "D")) {
    write !, tExternalName, " was not imported into the database and will not be compiled. "
} elseif (..ModifiedFiles(i).changeType = "D") {
    #; Deleted production items will not have an internal name
    set delIndex = $select(##class(SourceControl.Git.Util.Production).ItemIsPTD(tExternalName): 1, 1: 2)
    set:(delIndex=1) internalName = "-"_i
    set delList(delIndex, internalName) = tExternalName   ;; <<<  SUBSCRIPT error being thrown here
}
```

In my git diff, there were only these 2 deleted files:
```bash
    D    Demo.ProcessingIssueReason.CLS    cls/Demo/ProcessingIssueReason.cls
    D        data/sample/Failure_Reasons.csv
```

Neither item is a PTD, so `$select` returns 2.  However, that CSV does not have an internal name, so `internalName` ends up blank and we end up with `<SUBSCRIPT>`.

----
Version 2.15.0
$zv: IRIS for Windows (x86-64) 2025.1 (Build 223U) Tue Mar 11 2025 18:14:42 EDT

cambot:
Fixes #928:  the Incremental Load pull handler so that when applying commits results in deleting a file not managed by `embedded-git-config.json`, do not crash with a `<SUBSCRIPT>` error.

## Test Plan
Examples of files you might have in your repository that embedded git won't sync into Iris:
- readme/markdown files
- git special files (Message templates, .gitignore, etc..)
- CICD config/support files
- sample/seed data files

The easiest way to reproduce this issue is to create a branch that adds (commits) one or more of these files, then change branches to/from this branch.  (no need to merge into the main branch.)

You could also push this branch to a remote, reset your local to before this branch, then "Pull changes", but it's the same mechanism under the hood as "Check out branch"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Baseline Export now includes decomposed production items when production decomposition is enabled (#680)
+- Take Ownership option in source control menu allows you to take ownership of an item edited by another user (#926)
 
 ## Fixed
 - Fixed issue where Generated Files Read-only option didn't entirely work for files not in source control (#712)
--- a/cls/SourceControl/Git/Extension.cls
+++ b/cls/SourceControl/Git/Extension.cls
@@ -35,12 +35,14 @@ XData Menu
 <MenuItem Name="ExportForce" Save="101" />
 <MenuItem Name="Import" />
 <MenuItem Name="ImportForce" />
+<MenuItem Name="TakeOwnership" />
 </Menu>
 <Menu Name="%SourceContext" Type="1">
 <MenuItem Name="AddToSC" Save="100" />
 <MenuItem Name="RemoveFromSC"/>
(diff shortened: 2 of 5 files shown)


isc-pbarton (changes requested):
One nitpick here, otherwise looks great

PR #946, opened 2026-03-17
Fix adding and committing percent routines

### LINKED ISSUE #944: Server Source Control dropdown does not show "commit" option for INT/MAC routines starting with "%", but works through web UI
For int/mac routines starting with "%" character, I don't see the option for commit, even though I can do it through the web ui.
How do I isolate the cause for this?

Image

isc-pbarton:
Fixes #944 

Manual testing in the container:
- initialize git repo with ##class(SourceControl.Git.API).Configure()
- Add a routine mapping in the USER namespace of %pbarton* to USER database
- From Git Web UI, export the VS Code workspace and open in VS Code
- Add a new routine %pbartontest.mac and save
- Check source control menus. It should have options to remove and commit. Test removing and then adding to Git, then committing. View git history to confirm it was committed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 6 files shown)


isc-dchui (commented):
Looks good, just one thing

isc-dchui (approved):

PR #956, opened 2026-04-14
fix display of branch names with special characters

### LINKED ISSUE #914: Branch not displaying in GitUI
One of my team mates created a branch that included a pound sign: `feature/example-#123`.  The pound sign appeared to prevent the branch from showing up in the GitUI window under both the short list of local branches and the expanded list when you click "More..."

Fortunately, he was able to "recover" his work through Embedded git by switching into Expert mode and creating a new branch without a pound sign in the name.

----

Version: 2.14.0

isc-pbarton:
Fixes #914 

# Testing plan
docker compose exec iris iris session iris
do ##class(SourceControl.Git.API).Configure() to initialize a git repository
docker compose exec iris bash
cd /usr/irissys/mgr/repo/USER
git branch "invalid#branch"

load the web ui: http://localhost:52774/isc/studio/usertemplates/gitsourcecontrol/webuidriver.csp/USER/
- the branch list in the left panel should show a warning message that there is a branch with unsupported characters.
- I should be able to create new branches with the supported characters.
- I should not be able to create new branches with the unsupported characters, and I should get a descriptive error message if I try.

### DIFF
--- a/git-webui/release/share/git-webui/webui/js/git-webui.js
+++ b/git-webui/release/share/git-webui/webui/js/git-webui.js
@@ -439,7 +439,7 @@ webui.SideBarView = function(mainView, noEventHandlers) {
             searchBar.onkeyup = function(){ 
                 let branchCards = accordionDiv.getElementsByClassName("branch-card");
 
-                var filter = searchBar.value.toUpperCase().replaceAll('/', '-');
+                var filter = searchBar.value.toUpperCase().replaceAll('/', '-').replaceAll('#', '_');
 
                 for (let i = 0; i < branchCards.length; i++) {
                     let card = branchCards[i]
@@ -469,7 +469,7 @@ webui.SideBarView = function(mainView, noEventHandlers) {
             var cardDiv = $('<div class="card custom-card branch-card">').appendTo(accordionDiv)[0];
             if (id.indexOf("local-branches") > -1) {
                 // parses the output of git branch --verbose --verbose
-                var matches = /^\*?\s*([\w-.@&_\/]+)\s+([^\s]+)\s+(\[.*\])?.*/.exec(ref);
+                var matches = /^\*?\s*([\w-.@&_\/\#]+)\s+([^\s]+)\s+(\[.*\])?.*/.exec(ref);
                 if (!matches) {
                     continue;
                 }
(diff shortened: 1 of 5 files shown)


isc-pbarton:
CI failing I think because of changes in the intersystemsdc/iris-community:latest image. Fails to install TestCoverage. When I run locally on the same base image I get an error that no IPM registries are configured.

isc-dchui:
Hm I'm not fully convinced we should necessarily support the "#" character (or spaces) in branch names. I feel like as long as we support [this](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#naming-branches-and-tags) we're fine.

isc-pbarton:
After discussion with @isc-dchui, we decided that just supporting pound sign is kind of arbitrary, but we want to keep the web UI simple rather than supporting everything. Plan is:

- support smaller set of branch names matching [this default set](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#naming-branches-and-tags)
- add branch name validation when you create a new branch, both in web UI and source control menus
- exclude unsupported branch names from web UI, and show warning message if more branches exist with unsupported names. this keeps us reasonably compatible with other git clients

git-webui/release/share/git-webui/webui/js/git-webui.js
```diff
@@ -562,9 +580,17 @@ webui.SideBarView = function(mainView, noEventHandlers) {
...
     
         $("#btn_createList").click(function(e)
-        {   
+        {
             var refName = $('#newBranchName').val()
-    
+
+            // Validate branch name: only allow characters matching GitHub defaults
+            // Allowed: a-z, A-Z, 0-9, period, hyphen, underscore, forward slash
+            var branchNamePattern = /^[a-zA-Z0-9._\/-]+$/;
```
isc-dchui: Could we refactor the pattern into a central location to be referred to to avoid duplication everywhere?

isc-dchui (commented):
Just one thing and a question about testing. What kind of testing did you do for this change?

git-webui/release/share/git-webui/webui/js/git-webui.js
isc-pbarton: I've refactored the branch name pattern and added a testing plan in the PR description.

isc-dchui (approved):

PR #966, opened 2026-05-11
Validate that SSH key file path is not a directory

### LINKED ISSUE #943: Failure if SSH key file path is a directory
Say you are configuring a new repository against a remote. You misinterpret the "SSH key file path" setting as a directory the key will be generated in, rather than a full file path for the key. Configuration fails with an unhelpful error message. We should make it a helpful error message instead.

isc-pbarton:
## Description
Resolves #943

Adds validation to confirm the SSH key file path is not a directory when configuring Embedded Git, both through the Configure() prompt and the Settings page.

## Testing
In the development container:
- `do ##class(SourceControl.Git.API).Configure()` to go through the configuration prompts. When prompted for ssh key file path, put in `/usr/irissys/mgr`: the Configure prompt displays an informative error and quits out at the end. Try again with `/usr/irissys/mgr/test_key.pub`: at the end it prompts you to create the SSH key, and successfully does so.
- Visit the settings page at http://localhost:52774/isc/studio/usertemplates/gitsourcecontrol/gitprojectsettings.csp and test directory and non-directory values in the "SSH Private Key File" input.

### DIFF
--- a/csp/gitprojectsettings.csp
+++ b/csp/gitprojectsettings.csp
@@ -384,7 +384,7 @@ body {
         </div>
 
         <server>
-        if (settings.privateKeyFile '= "") && fileExists && '$$$ISERR(validateSC) {
+        if (settings.privateKeyFile '= "") && fileExists && $$$ISOK(validateSC) {
             set pubKeyName = settings.privateKeyFile_".pub"
             if ##class(%File).Exists(pubKeyName) {
                 set pubStream = ##class(%Stream.FileCharacter).%OpenId(pubKeyName)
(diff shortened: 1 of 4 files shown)


csp/gitprojectsettings.csp
```diff
@@ -379,7 +384,7 @@ body {
         </div>
 
         <server>
-        if (settings.privateKeyFile '= "") && fileExists {
+        if (settings.privateKeyFile '= "") && fileExists && '$$$ISERR(validateSC) {
```
isc-dchui: What's wrong with $$$ISOK()?

isc-pbarton: I fixed, retested, and resolved conflicts.

isc-dchui (approved):

PR #972, opened 2026-06-04
Fix invalid regular expression in branch name special character check

raymond-rebbeck:
## Description
Tried latest main branch on Studio and noticed there was an error when opening Git UI about invalid quantifier pointing at line 477 in `git-webui.js` - likely caused by the initial `\*` not being escaped and being treated as a regex operator.

After fixing this did not get any local branches listed in Studio or VS Code. After some investigating determined the following:
- Backslash characters not correctly escaped.
- `webui.branchNamePattern.source` starts and ends with `^` and `$` respectively, which do not work when inserted into the middle of another regular expression so chop off the first and last characters to avoid this.

## Testing
Tested locally on my own IRIS instance with several local branches. With VS Code and Studio.

## Checklist
- [x] This branch has the latest changes from the `main` branch rebased or merged.
- [x] Web UI has been built (any changes in `git-webui/src` have matching changes in `git-webui/release`)
- [N/A] CHANGELOG.md entry added if appropriate.

### DIFF
--- a/git-webui/release/share/git-webui/webui/js/git-webui.js
+++ b/git-webui/release/share/git-webui/webui/js/git-webui.js
@@ -472,9 +472,18 @@ webui.SideBarView = function(mainView, noEventHandlers) {
             }
             var cardDiv = $('<div class="card custom-card branch-card">').appendTo(accordionDiv)[0];
             if (id.indexOf("local-branches") > -1) {
+                // Remove any '^' prefix or '$' suffix from the branch name regex
+                // So that it can be inserted into the middle of another regex below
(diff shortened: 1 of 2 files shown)


raymond-rebbeck:
Just updated this so that it will check if there is a '^' prefix or '$' suffix on the branch name regex before removing them, rather than just always blindly removing the first and last characters.

isc-pbarton (approved):

isc-pbarton:
Approved. CI failure is unrelated, and changelog not needed as this is a fix to an unreleased feature (#914)
Thank you for the contribution!

PR #976, opened 2026-06-12
Fix CI by using new JUnit output argument for IPM

isc-pbarton:
Some problem with IPM 0.10.7 where the JUnit test output is not being created. I will log an IPM issue as well and we can go back to latest if that is fixed
Edit: gone back to latest, fixed by using the new -output-file argument

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,7 +64,7 @@ jobs:
 
       - name: Install IPM
         run: |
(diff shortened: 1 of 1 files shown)


isc-pbarton:
IPM bug report here: https://github.com/intersystems/ipm/issues/1196

isc-dchui (approved):

PR #995, opened 2026-08-19
When adding a record map to source control, automatically add generated classes

### LINKED ISSUE #955: Adding a RecordMap class to Git should automatically add the corresponding data class
Saving/compiling a Record mapper class does not automatically add it to git, so I had to add it manually (Other compiled assets like DTLs and BPLs act the same way, so I can roll with this).  However, when I added my mapper pool class, I also had to manually add the generated data class.

These record mapper classes extend `EnsLib.RecordMap.RecordMap`, which has a parameter to define the target message/object class.  It would be a great feature for Embedded git to:
1. When adding the RecordMap class, also add this object class
2. when compiling/saving/exporting the RecordMap class, also export the object class. 

Developers are inevitably going to miss manually adding the object class.

isc-pbarton:
## Description
Resolves #955
Adding a record map to source control adds its associated data classes to source control. 
Separately, creating a record map will automatically add it to source control.

_why source control the generated classes?_
They are not automatically generated on compilation. Generation of the data classes is a manual step. If the data classes were not source controlled, there would need to be another manual step on deployment to another environment.

_why can't we just reuse the logic in the deployment manager to find related classes?_
The deployment manager has some nice logic that will find all classes required by a specific business host so they can be added to a deployment. We originally thought we could use that, and make this more generic than just record maps. But that's not a good fit for source control hooks on record maps because we can't assume a business host exists using the record map.

## Testing
Unit tests, plus some manual testing on record maps:
- configure the IRIS container with a git repository
- create a new record map in the SMP and click "generate"
- check source control output to confirm both record map and data class have been exported.

## Checklist
- [x] This branch has the latest changes from the `main` branch rebased or merged.
- [N/A] Web UI has been built (any changes in `git-webui/src` have matching changes in `git-webui/release`)
- [x] CHANGELOG.md entry added if appropriate.


cls/SourceControl/Git/Extension.cls
```diff
@@ -64,10 +64,10 @@ Method UserAction(Type As %Integer, Name As %String, InternalName As %String, Se
         do ##class(SourceControl.Git.Change).GetUncommitted(filename,.tAction)
         do ..GetStatus(.InternalName, .isInSourceControl, .isEditable,.isCheckedOut,.userCheckedOut)
 
-        // Deal with Business Processes and Rules
-        // Note: Business Processes and Rules do not have a 'new document' User Action, and thus must be added like this
+        // certain portal UIs create and compile classes without going through the 'new document' user
```
isc-dchui, line 67: Nit: capitalization

isc-dchui (approved):
Just one tiny nit

PR #1003, opened 2026-09-02
Better handling of IPM packages mapped from another namespace

### LINKED ISSUE #997: Embedded Git deletes itself on import
Issue experienced with a customer:
Running Import All would delete SourceControl.Git.Utils class, thus breaking the source control extension.

Immediate cause was `^SYS("SourceControl","Git","TSH","SourceControl.Git.Utils.CLS")` global subnode was set, and workaround was to kill it. Not entirely sure why they were set.

The bug here is that Import All should never be deleting things if they are from a package manager context.

isc-pbarton:
Resolves #997 
This works, but feels far too heavyweight and I'm uncertain of the performance impacts. Saving this work against a draft PR, and will put up a separate PR with a more targeted fix to compare.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Record map classes and their generated data classes are now automatically added to source control (#955)
 
+### Fixed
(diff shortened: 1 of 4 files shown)


isc-pbarton:
A better design is specified in issue #1006

